PR #78, opened 2025-10-27
hash truncation

### LINKED ISSUE #53: [core] Add global hash truncation configuration
Add support for globally truncating all hash outputs (data versions, feature versions, snapshot versions) to reduce storage and improve readability.

Hash truncation should be configured at the Metaxy level, not per-store, to ensure consistency across all hash types.

## Hash Types Affected
- `data_version` / `provenance_by_field` - computed in DataVersionCalculator
- `feature_version` - computed in Feature class
- `snapshot_version` - computed in FeatureGraph
- Migration hashes (if any)

## Acceptance Criteria
- [x] Add global configuration option for hash truncation (metaxy.toml or environment variable)
- [x] Apply truncation consistently to ALL hash computations
- [x] Implement in Feature.feature_version() method
- [x] Implement in FeatureGraph.snapshot_version property
- [x] Implement in DataVersionCalculator implementations
- [x] Default to no truncation (full hash) for backward compatibility
- [x] Add validation for minimum truncation length (e.g., min 8 chars)
- [x] Ensure all stores in a deployment use same truncation
- [x] Add comprehensive tests:
  - [x] Test truncation in Feature.feature_version()
  - [x] Test truncation in FeatureGraph.snapshot_version
  - [ ] Test truncation in DataVersionCalculator for each implementation
  - [x] Test configuration loading from metaxy.toml
  - [x] Test environment variable override
  - [x] Test validation of minimum length
  - [ ] Test that changing truncation is detected as breaking change
- [ ] Document collision probability and recommended lengths

## Configuration Options
```toml
# metaxy.toml
[core]
hash_truncate_length = 16  # Truncate all hashes to 16 chars

# OR via environment variable
METAXY_CORE_HASH_TRUNCATE_LENGTH=16
```

## Considerations
- Truncation must be consistent across entire deployment
- Changing truncation length is a breaking change (requires migration). How would we handle such migrations? Also, how would we handle migrations between hash algorithms? Changing it won't create a new snapshot ID, so existing data version reconciliation migration op would need slight modifications.
- Consider warning if length seems too short for scale
- 16 chars of hex = 64 bits = ~10^19 possible values (very low collision risk)

danielgafni:
Resolve #53

### DIFF
--- a/src/metaxy/utils/hashing.py
+++ b/src/metaxy/utils/hashing.py
@@ -1,0 +1,205 @@
+"""Hash truncation utilities for Metaxy.
+
+This module provides utilities for globally truncating hash outputs to reduce
+storage requirements and improve readability. Hash truncation is configured
+through the global MetaxyConfig.
+"""
+
+from typing import Any
+
+import narwhals as nw
+
+# Minimum allowed truncation length
+MIN_TRUNCATION_LENGTH = 8
+
+
+def truncate_hash(hash_str: str) -> str:
+    """Truncate a hash string using the global truncation setting.
+
+    Uses the global hash truncation setting from MetaxyConfig.
+    If the global setting is None, returns the full hash.
+
+    Args:
+        hash_str: The hash string to truncate
+
+    Returns:
+        Truncated hash string
+
+    Examples:
+        >>> # With global config set to truncation_length=12:
+        >>> truncate_hash("a" * 64)
+        'aaaaaaaaaaaa'
+
+        >>> # With no truncation setting:
+        >>> truncate_hash("abc123")
+        'abc123'
+    """
+    # Get length from global setting
+    length = get_hash_truncation_length()
+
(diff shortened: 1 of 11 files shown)


src/metaxy/utils/hashing.py
```diff
@@ -0,0 +1,142 @@
...
+This module provides utilities for globally truncating hash outputs to reduce
+storage requirements and improve readability. Hash truncation is configured
+through the global MetaxyConfig.
+"""
+
+# Minimum allowed truncation length
+MIN_TRUNCATION_LENGTH = 8
+
+
+def truncate_hash(hash_str: str, length: int | None = None) -> str:
```
danielgafni: this should not take length as an option, length be taken with `get_hash_truncation_length` 

also, a narwhals (`@narwhalify`) function doing the same with
- a column
- a struct (truncate values)

is needed

Issue #62, opened 2025-10-27
[core] Change FieldSpec code_version from int to string


danielgafni:
## Description
Currently, `FieldSpec` uses an integer for the `code_version` parameter. This should be changed to a string to provide more flexibility and better semantic versioning capabilities.

## Motivation
- Integer versions are limiting and don't allow for semantic versioning (e.g., "1.2.3", "v2.0.0-beta")
- String versions can include git commit hashes for better traceability
- More consistent with industry standards for version identifiers
- Allows for more descriptive version identifiers when needed

## Acceptance Criteria
- [ ] Change `FieldSpec.code_version` type from `int` to `str`
- [ ] Update all existing usages of `code_version` throughout the codebase
- [ ] Update type annotations and documentation
- [ ] Ensure backward compatibility for migration from int to string versions
- [ ] Update tests to use string versions
- [ ] Verify that version hashing still works correctly with string versions

## Implementation Notes
- This is a breaking change but acceptable since we have no users yet
- Consider providing a migration path or automatic conversion for existing int versions
- Ensure that version comparison logic (if any) is updated to handle string versions
- The feature version hash calculation should continue to work with string code_versions

## Files to Update
Key files likely affected:
- `src/metaxy/models/feature_spec.py` - FieldSpec definition
- Any files using `FieldSpec(code_version=...)`
- Tests that create FieldSpec instances
- Documentation mentioning code_version

## Example Usage
Before:
```python
FieldSpec(
    name="my_field",
    code_version=1
)
```

After:
```python
FieldSpec(
    name="my_field", 
    code_version="1.0.0"  # or "v1", "2024.1", "abc123", etc.
)
```

geoHeil:
new version after various breaking api changes https://github.com/anam-org/metaxy/pull/183

PR #84, opened 2025-10-28
Change FieldSpec code_version from int to string #62

geoHeil:


danielgafni:
Uhh I'm very sorry, I missed this one. It has to be rebased frequently because code versions are all over the place. 

Actually I propose to include this change into our API overhaul, and make it backwards-compatible. 

Or make your current PR backwards compatible. 

I don't see issues with accepting integers as long as they are converted to strings anyway.

geoHeil:
superseeded by https://github.com/anam-org/metaxy/pull/104

Issue #71, opened 2025-10-27
[core] Accept more types into `FeatureSpec` parameters


danielgafni:
## Description
Currently, FeatureSpec and FieldSpec require explicit FeatureKey and FieldKey wrappers for all key parameters. This adds unnecessary verbosity to feature definitions.

## Problem
```python
# Current - verbose
class MyFeature(Feature, spec=FeatureSpec(
    key=FeatureKey(["my_feature"]),
    deps=[FeatureDep(key=FeatureKey(["upstream"]))],
    fields=[FieldSpec(key=FieldKey(["result"]))]
)):
    pass
```

## Proposed Solution
Accept Python native types (strings, lists) and Feature classes directly:

```python
# Proposed - concise
class MyFeature(Feature, spec=FeatureSpec(
    key="my_feature",  # Single string
    deps=[UpstreamFeature],  # Feature class directly
    fields=[
        FieldSpec(key="result"),  # String for single-part key
        FieldSpec(key=["nested", "field"])  # List for multi-part key
    ]
)):
    pass
```

## Supported Type Conversions
- `str` → `FeatureKey([str])` or `FieldKey([str])`
- `list[str]` → `FeatureKey(list)` or `FieldKey(list)`  
- `type[Feature]` → `FeatureDep(key=Feature.spec.key)`
- `FeatureSpec` → `FeatureDep(key=spec.key)`

## Acceptance Criteria
- [ ] FeatureSpec.key accepts `str | list[str] | FeatureKey`
- [ ] FieldSpec.key accepts `str | list[str] | FieldKey`
- [ ] FeatureDep accepts `type[Feature] | FeatureSpec | FeatureKey | str | list[str]`
- [ ] FieldDep.feature_key accepts all FeatureKey-compatible types
- [ ] Pydantic validators handle type coercion
- [ ] Full backward compatibility with existing code

## Related
Part of #70 (Improve Feature and FeatureSpec API ergonomics)

danielgafni:
@geoHeil the right way to do it so to override __new__ - if it plays well with Pydantic

geoHeil:
What about

```
@field_validator("key", mode="before")
    @classmethod
    def _coerce_key(
```

https://github.com/anam-org/metaxy/pull/103/files

danielgafni:
Actually this works as I wanted it to: 

```py


from pydantic import BaseModel, field_validator
from typing_extensions import Any, overload, reveal_type


class MyModel(BaseModel):
    foo: str

[...]

PR #103, opened 2025-10-29
#71 Accept more types into FeatureSpec parameters

geoHeil:
https://github.com/anam-org/metaxy/issues/71


danielgafni:
This is not what we want, we can't validate the expected `Union` against type checkers, right? 
This needs a full refactor Pydantic is not usable as user API frontend, hold back on this feature for now.

Issue #72, opened 2025-10-27
[core] Support logical features


danielgafni:
**Logical features support** - Allow features with zero fields to act as dependency groupings.

Logical features inherit all parent fields under their feature keys namespaces. 
Features that depend on logical feature fields can "passs through" to their parents.
This enables scenarios like one-to-many mapping and others. 

 ```python
 class Video(Feature, spec=FeatureSpec(
     key=["video"], fields=["frames", "audio"]
 )):
     pass

   class VideoSplittingState(Feature, spec=FeatureSpec(
     deps=[Video], logical=True  # fields automatically populated to: [["video/frames", "video/audio"]]
 )):
     pass

  class AllInputs(Feature, spec=FeatureSpec(
     deps=[VideoSplittingState],   # Video is added as a dependency as well automatically
     fields=[...]  # fields can refer to video/frames field of VideoSplittingState, dependencies on corresponding Video are created automatically
 )):
     pass

 ```

There is a lot here to think about. 

Seems like it would naturally resolve https://github.com/anam-org/metaxy/issues/65 (logical features will be hidden)

danielgafni:
That's a bunch of nonsense, I regret writing this.

It's trivial to just do 1->many mapping directly, we just need to implement ID mapping for that. That's trivial to do.

PR #104, opened 2025-10-29
#72 Change FieldSpec code_version from int to string

geoHeil:
https://github.com/anam-org/metaxy/issues/62


geoHeil:
superseeded by https://github.com/anam-org/metaxy/pull/120

PR #111, opened 2025-10-29
#108 support python 3.14

### LINKED ISSUE #108: 3.14 support
Update CI and library support for python version 3.14

geoHeil:
Resolve #108


danielgafni:
Yes, go for it. I'm not sure if that's the problem here, but you can try. I won't focus on this right now myself.

PR #118, opened 2025-10-29
Python 3.14 support

geoHeil:
https://github.com/anam-org/metaxy/issues/108

### DIFF
--- a/.github/workflows/QA.yml
+++ b/.github/workflows/QA.yml
@@ -79,7 +79,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.10", "3.11", "3.12", "3.13"]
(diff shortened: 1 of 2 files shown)


.github/workflows/QA.yml
```diff
@@ -79,7 +79,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.10", "3.11", "3.12", "3.13"]
+        python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
```
geoHeil: watchdog 6.0.0 doesn’t (yet) declare/support Python 3.14, and its sdist for macOS bundles a prebuilt *_cpython-310-darwin.so, so when building on 3.14 the wheel ends up with mismatched tags

geoHeil: https://pypi.org/project/watchdog/#files

PR #171, opened 2025-11-01
Refine telemetry & docs: custom DNS

geoHeil:


mkdocs.yml
```diff
@@ -1,6 +1,6 @@
 site_name: Metaxy
 site_description: Metadata layer for multi-modal ML pipelines
-site_url: https://anam-org.github.io/metaxy
+site_url: https://docs.metaxy.io
```
geoHeil: only merge once the DNS check and GH-pages settings are OK and changed to `docs.metaxy.io`

PR #183, opened 2025-11-02
feat: int to str for code version

### LINKED ISSUE #62: [core] Change FieldSpec code_version from int to string

geoHeil:
resolves #62

### DIFF
--- a/src/metaxy/graph/diff/diff_models.py
+++ b/src/metaxy/graph/diff/diff_models.py
@@ -17,8 +17,8 @@ class FieldChange(FrozenBaseModel):
     field_key: FieldKey
     old_version: str | None = None  # None if field was added
     new_version: str | None = None  # None if field was removed
-    old_code_version: int | None = None
-    new_code_version: int | None = None
+    old_code_version: str | None = None
+    new_code_version: str | None = None
 
     @property
     def is_added(self) -> bool:
@@ -46,8 +46,8 @@ class NodeChange(FrozenBaseModel):
     feature_key: FeatureKey
     old_version: str | None = None  # None if node was added
     new_version: str | None = None  # None if node was removed
-    old_code_version: int | None = None
-    new_code_version: int | None = None
+    old_code_version: str | None = None
+    new_code_version: str | None = None
     added_fields: list[FieldChange] = Field(default_factory=list)
     removed_fields: list[FieldChange] = Field(default_factory=list)
     changed_fields: list[FieldChange] = Field(default_factory=list)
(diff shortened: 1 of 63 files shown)


src/metaxy/graph/diff/diff_models.py
```diff
@@ -201,7 +201,7 @@ def to_struct(self) -> dict[str, Any]:
                     {
                         "key": field.field_key.to_string(),
                         "version": field.old_version or "",
-                        "code_version": field.old_code_version or 0,
+                        "code_version": field.old_code_version or "",
```
danielgafni: This is definitely wrong, we should never have an empty code version. 

It should default to a known shared constant like "0". 

More importantly, **we should not randomly assume the default code version around the codebase** in places **where it should already be set at this point**. 

If a dictionary that **should have it** doesn't actually have this field, we should just fail naturally, because this means something went really wrong in the previous code paths that should have been validating this invariant.

danielgafni (changes requested):
We should establish a shared constant for the default code version and only use it once (or in very few places if absolutely necessary, but I suspect there should be no other way to set a code version other than through FieldSpec): during FieldSpec construction. 

We should eliminate all the dangerous defaults in the codebase (unfortunately AI really likes writing such code).

danielgafni:
Since was already a flaw of the codebase - I can see fixing it as part of another PR up the stack

danielgafni (approved):
Yes actually let's do it upstack, but it must be done.

PR #185, opened 2025-11-02
fix: forbid missing code_version in various code paths 

### LINKED ISSUE #184: [core] fix: never have no code version
```python
"code_version": field.old_code_version or 0,
"code_version": field.old_code_version or "",
```

This is definitely wrong, we should never have an empty code version.

It should default to a known shared constant like "0".

More importantly, we should not randomly assume the default code version around the codebase in places where it should already be set at this point.

If a dictionary that should have it doesn't actually have this field, we should just fail naturally, because this means something went really wrong in the previous code paths that should have been validating this invariant.

geoHeil:
resolves #184

### DIFF
--- a/tests/graph/test_struct_serialization.py
+++ b/tests/graph/test_struct_serialization.py
@@ -1,5 +1,7 @@
 """Tests for GraphData and GraphDiff struct serialization."""
 
+import pytest
+
 from metaxy.graph.diff.diff_models import (
     AddedNode,
     FieldChange,
@@ -9,6 +11,7 @@ from metaxy.graph.diff.diff_models import (
 )
 from metaxy.graph.diff.models import EdgeData, FieldNode, GraphData, GraphNode
 from metaxy.models.types import FeatureKey, FieldKey
+from metaxy.utils.exceptions import MetaxyEmptyCodeVersionError
 
 
 def test_graphdata_to_struct():
@@ -372,11 +375,8 @@ def test_graphdata_handles_none_values():
         edges=[],
     )
 
-    struct = graph_data.to_struct()
-    restored = GraphData.from_struct(struct)
-
-    node = restored.nodes["feature/a"]
-    assert node.version is None
-    assert node.code_version is None
-    assert node.fields[0].version is None
-    assert node.fields[0].code_version is None
+    with pytest.raises(
+        MetaxyEmptyCodeVersionError,
+        match="Field field1 in feature feature/a has empty code_version",
+    ):
+        graph_data.to_struct()


src/metaxy/graph/diff/models.py
```diff
@@ -13,6 +13,11 @@
     from metaxy.models.feature import FeatureGraph
 
 
+# Sentinel value for None code_version in struct serialization
+# (Polars structs require a non-None value, so we use "0" to represent missing/None)
+EMPTY_CODE_VERSION = "0"
```
geoHeil: and this is something we should move over to a shared utils so it is not re-defined

src/metaxy/graph/diff/diff_models.py
```diff
@@ -145,15 +149,15 @@ def to_struct(self) -> dict[str, Any]:
...
+                        "code_version": field["code_version"],
                     }
                 )
 
             added_nodes_list.append(
                 {
                     "key": node.feature_key.to_string(),
                     "version": node.version,
-                    "code_version": node.code_version or "",
+                    "code_version": node.code_version or EMPTY_CODE_VERSION,
```
geoHeil: Wouldn`t we want to reaise here as well?

src/metaxy/graph/diff/models.py
danielgafni: Rename to DEFAULT_CODE_VERSION (it's not empty!) and let's set it to "__metaxy__intitial__"

danielgafni (changes requested):
Minor comments. 

We should raise everywhere.

Yes the only way a code version can get into our system is through FieldSpec. 

Maybe make an InvariantViolationError like Dagster has and explain this.

src/metaxy/graph/diff/diff_models.py
geoHeil: done

danielgafni:
Nice! 

A future PR could be consolidating all the "code_version" accesses to use the same key: METAXY_CODE_VERSION_KEY.

I'd like to consolidate all magic column names that we are using in order to rename all of them with metaxy_ prefix.

danielgafni:
The failing test should be fixed to actually expect our error to be raised (pytest.raises)

geoHeil:
> Nice!
> 
> A future PR could be consolidating all the "code_version" accesses to use the same key: METAXY_CODE_VERSION_KEY.
> 
> I'd like to consolidate all magic column names that we are using in order to rename all of them with metaxy_ prefix.

https://github.com/anam-org/metaxy/issues/186 lets tackle that there

PR #189, opened 2025-11-02
[core] remove Feature.code_version() classmethod

### LINKED ISSUE #168: [streamline API] remove code_version from feature and only keep it for feature_spec
## Summary

  Keep `code_version` on `FeatureSpec` only, remove from `Feature` API surface to improve clarity and reduce API pollution.

  ## Motivation

  Currently, `code_version` can be accessed via `Feature`, but there are good reasons to keep it exclusively on `FeatureSpec`:

  1. **API surface cleanliness**: Users interact with the `Feature` class constantly and mostly expect to see their own fields on it. We don't want to pollute
  the API with framework internals.

  2. **Precedent**: We already removed `Feature.id_columns()` for the same reason - to keep Feature's API focused on user concerns.

  3. **Clarity about what's versioned**: Without reading the docs, it's not obvious that this property simply forwards data from `FeatureSpec`. By keeping it
  on `FeatureSpec` only, we send a clear signal that the Feature itself is not involved in this version in any way - it's purely about the specification.

  ## Proposed Changes

  - Remove `Feature.code_version` property/accessor from `src/metaxy/models/feature.py`
  - Update any internal code that currently uses `Feature.code_version` to use `feature.spec.code_version` instead
  - Update documentation to reflect this API design decision

  ## Acceptance Criteria

  - [ ] `Feature.code_version` property/accessor removed
  - [ ] All internal references updated to use `feature.spec.code_version`
  - [ ] Tests pass
  - [ ] Documentation updated

geoHeil:
resolve #168

### DIFF
--- a/docs/reference/api/definitions/feature-spec.md
+++ b/docs/reference/api/definitions/feature-spec.md
@@ -2,9 +2,6 @@
 
 Feature specs act as source of truth for all metadata related to features: their dependencies, fields, code versions, and so on.
 
-!!! tip "Accessing Code Versions"
-    Use `MyFeature.spec().code_version` to inspect the deterministic hash for a feature definition. This value is only exposed via the spec to keep the `Feature` API focused on user-defined fields.
-
 ::: metaxy.BaseFeatureSpec
 
 ::: metaxy.FeatureSpec


docs/reference/api/definitions/feature.md
```diff
@@ -9,3 +9,6 @@ Users can extend this class to define their features.
 ::: metaxy.BaseFeature
 
 ::: metaxy.Feature
+
+!!! note "Code Version Access"
```
danielgafni: Let's put this in between BaseFeature and Feature. 

Also, let's used > ! notation for consistency with GitHub markdown flavor

PR #190, opened 2025-11-02
feat: no pydantic

### LINKED ISSUE #179: [core] Reassess whether BaseFeature needs to be a Pydantic model
It probably doesn't since we never serialize it. We only serialize FeatureSpec.

geoHeil:
resolves #179

### DIFF
--- a/examples/src/examples/overview/features_1.py
+++ b/examples/src/examples/overview/features_1.py
@@ -1,3 +1,5 @@
+from typing import ClassVar
+
 from metaxy import (
     Feature,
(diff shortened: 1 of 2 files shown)


danielgafni (commented):
After thinking about this again, once we consolidate `metaxy_` metadata columns on the `BaseFeature`, we would start to benefit from Pydantic's fields more. 

Also, Pydantic's field registry would act as a known location for user-defined fiels. 

So I think we should keep `BaseFeature` to be a Pydantic model.

PR #218, opened 2025-11-04
Add customizable table naming logic to Ibis stores

### LINKED ISSUE #51: [core] Add customizable table naming logic to Ibis stores
Add support for customizable table naming in Ibis-based metadata stores, allowing users to define their own naming logic beyond simple prefixes.

## Acceptance Criteria
- [ ] Add user-facing method for table name generation (e.g., `get_table_name()`)
- [ ] Support both simple prefix and custom naming logic
- [ ] Default implementation uses prefix if provided
- [ ] Allow users to override/customize the naming method
- [ ] Apply to all feature and system table names
- [ ] Ensure names are validated (valid SQL identifiers)
- [ ] Update all Ibis store implementations
- [ ] Add tests for custom naming scenarios
- [ ] Document how to implement custom naming strategies

## Example Usage
```python
# Simple prefix
store = DuckDBMetadataStore(
    connection_string="my_database.db",
    table_prefix="prod_v2_"
)

# Custom naming logic
class MyCustomStore(DuckDBMetadataStore):
    def get_table_name(self, feature_key: str, is_system: bool = False) -> str:
        """User-facing method to customize table naming."""
        if is_system:
            return f"sys_{feature_key}_{self.environment}"
        return f"{self.namespace}_{feature_key}_v{self.version}"

store = MyCustomStore(
    connection_string="my_database.db",
    namespace="ml",
    version=2,
    environment="prod"
)
# Creates tables like: ml_my_feature_v2, sys_migrations_prod
```

geoHeil:
resolves #51

### DIFF
--- a/src/metaxy/graph/diff/differ.py
+++ b/src/metaxy/graph/diff/differ.py
@@ -674,9 +674,7 @@ class GraphDiffer:
                 )
 
         except Exception as e:
-            raise ValueError(
-                f"Failed to load snapshot {snapshot_version}: cannot load snapshot {snapshot_version}: {e}"
-            ) from e
+            raise ValueError(f"Failed to load snapshot {snapshot_version}: {e}") from e
 
         # Build snapshot data structure
         snapshot_dict = {}
--- a/src/metaxy/metadata_store/ibis.py
+++ b/src/metaxy/metadata_store/ibis.py
@@ -367,7 +367,7 @@ class IbisMetadataStore(MetadataStore, ABC):
 
             df_to_insert = collect_to_polars(df)  # Polars DataFrame
 
-        table_name = feature_key.table_name
+        table_name = self.get_table_name(feature_key)
 
         try:
             self.conn.insert(table_name, obj=df_to_insert)  # type: ignore[attr-defined]  # pyright: ignore[reportAttributeAccessIssue]
--- a/src/metaxy/migrations/executor.py
+++ b/src/metaxy/migrations/executor.py
@@ -4,7 +4,6 @@ This is the new executor that replaces the old 3-table system with a single
 event-based system stored in system tables via SystemTableStorage.
 """
 
-import traceback
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
@@ -246,22 +245,21 @@ class MigrationExecutor:
                     rows_affected_total += rows_affected
 
                 except Exception as e:
-                    # Capture both human-friendly and full traceback messages
-                    short_error = str(e) if str(e) else repr(e)
-                    trace_error = traceback.format_exc().strip()
-                    errors[feature_key_str] = f"{short_error}\n{trace_error}"
+                    # Get full error message
+                    error_msg = str(e) if str(e) else repr(e)
+                    errors[feature_key_str] = error_msg
 
                     # Log exception with full traceback
                     logger.exception(f"Error in feature {feature_key_str}")
 
-                    # Log feature failed with concise error in system tables
+                    # Log feature failed
                     if not dry_run:
                         self.storage.write_event(
                             Event.feature_failed(
                                 project=project,
                                 migration_id=migration.migration_id,
                                 feature_key=feature_key_str,
-                                error_message=short_error,
+                                error_message=error_msg,
                             )
                         )
 
@@ -442,9 +440,9 @@ class MigrationExecutor:
                     rows_affected_total += rows_affected
 
                 except Exception as e:
-                    short_error = str(e) if str(e) else repr(e)
-                    trace_error = traceback.format_exc().strip()
-                    errors[feature_key_str] = f"{short_error}\n{trace_error}"
+                    # Get full error message
+                    error_msg = str(e) if str(e) else repr(e)
+                    errors[feature_key_str] = error_msg
 
                     # Log exception with full traceback
                     logger.exception(f"Error in feature {feature_key_str}")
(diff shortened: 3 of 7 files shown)


danielgafni (changes requested):
This is supposed to be a tiny change only touching ibis stores constructor, the table name method, and adding one unit test for the modified method. Other stores call this method directly so there is no need to test them for this behavior. 

There is no need to do any docs changes at all. Users will read about this parameter in the API docs.

geoHeil:
> This is supposed to be a tiny change only touching ibis stores constructor, the table name method, and adding one unit test for the modified method. Other stores call this method directly so there is no need to test them for this behavior.
> 
> There is no need to do any docs changes at all. Users will read about this parameter in the API docs.

ok we can remove docs, tests perhaps but I would also suggest to change from - to _ and I think then the change is a bit larger. Why do you think this is not something which should be explained in the docs?

danielgafni (changes requested):
I think we are almost there!

src/metaxy/metadata_store/ibis.py
```diff
@@ -117,9 +121,52 @@ def __init__(
         self.backend = backend
         self.connection_params = connection_params or {}
         self._conn: ibis.BaseBackend | None = None
+        self._table_prefix = table_prefix or ""
 
         super().__init__(**kwargs)
 
+    def get_table_name(
+        self,
+        key: FeatureKey | Sequence[str] | str,
```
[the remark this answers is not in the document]
geoHeil: vote for or actually have the system constants as real FeatureKey instances - this would make more sense.

https://github.com/anam-org/metaxy/issues/276

src/metaxy/metadata_store/ibis.py
```diff
@@ -77,6 +78,9 @@ def __init__(
                 Used with connection_params for more control.
             connection_params: Backend-specific connection parameters
                 e.g., {"host": "localhost", "port": 9000, "database": "default"}
+            table_prefix: Optional prefix applied to all feature and system table names.
+                Useful for isolating environments (e.g., "prod_"). Must form a valid SQL
```
danielgafni: I guess not "isolating" but "logically separating"

danielgafni (approved):
Nice! Approved with a minor comment

geoHeil:
https://github.com/anam-org/metaxy/pull/218#discussion_r2523755988 in the list table we are literally listing the tables and deriving features from the table names 


we do not have a metadata store somewhere hosting pointers to features - as a replacement we scan all tables in the featuregraph


but you mean - we should be able to construct the graph from the yaml files/configuration? instead of from the tables?

danielgafni (changes requested):
I propose the following: 

1. Drop all changes regarding `list_features`​. Do not touch these code paths. This PR should only change the table naming logic. It's backcompat, so no tests should break, right? 
2. After that rebase (`gt move`​) on top of `11-12-final_pass_on_migrations`​ - I removed `list_features`​ and fixed all tests as part of this PR. Your PR should have no (big) conflicts and continue to work.

src/metaxy/cli/metadata.py
```diff
@@ -253,11 +253,8 @@ def drop(
     with metadata_store.open(AccessMode.WRITE):
         # If all_features, get all feature keys from the active feature graph
         if all_features:
-            from metaxy.models.feature import FeatureGraph
-
-            graph = FeatureGraph.get_active()
-            # Get all feature keys from the graph (features defined in code for current project)
-            feature_keys = graph.list_features(only_current_project=True)
+            # Use only features defined in the active graph for the current project
+            feature_keys = metadata_store.list_features()
```
danielgafni: Why change this?

src/metaxy/metadata_store/ibis.py
```diff
@@ -206,6 +236,27 @@ def _validate_hash_algorithm_support(self) -> None:
                 f"Supported algorithms: {', '.join(supported)}"
             )
 
+    def _ensure_system_tables_exist(self) -> None:
```
danielgafni: This should be dropped, the normal metadata write method already creates tables if auto creation is enabled

src/metaxy/cli/metadata.py
geoHeil: resetted

src/metaxy/migrations/executor.py
```diff
@@ -245,21 +246,22 @@ def _execute_diff_migration(
                     rows_affected_total += rows_affected
 
                 except Exception as e:
-                    # Get full error message
-                    error_msg = str(e) if str(e) else repr(e)
-                    errors[feature_key_str] = error_msg
+                    # Capture both human-friendly and full traceback messages
+                    short_error = str(e) if str(e) else repr(e)
+                    trace_error = traceback.format_exc().strip()
+                    errors[feature_key_str] = trace_error
```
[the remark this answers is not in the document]
geoHeil: fixed

src/metaxy/graph/diff/differ.py
```diff
@@ -674,7 +674,9 @@ def load_snapshot_data(
                 )
 
         except Exception as e:
-            raise ValueError(f"Failed to load snapshot {snapshot_version}: {e}") from e
+            raise ValueError(
```
danielgafni: irrelevant

tests/metadata_stores/test_ducklake.py
danielgafni: This diff is very confusing, why are we changing existing tests?

danielgafni: Well, the test is passing/skipping now, so why touch it? 

I understand you fixed it, but this created additional overhead for me: k want to focus on the actual useful changes that this PR is supposed to introduce. I don't want to be looking at diffs that are not related. 

Extra stuff is causing friction and prevents the main functionality from getting merged. You can do these changes in separate PRs without blocking yourself, that's what Graphite is for.

PR #221, opened 2025-11-04
Add METAXY_AUTO_PUSH_GRAPH environment variable

### LINKED ISSUE #56: [core] Add METAXY_AUTO_PUSH_GRAPH environment variable
Add automatic graph snapshot recording when metadata stores are opened.

When enabled, stores will call `record_feature_graph_snapshot()` on open to ensure the feature_versions table stays current.

## Acceptance Criteria
- [ ] Add `METAXY_AUTO_PUSH_GRAPH` environment variable
- [ ] Add `[core] auto_push_graph` option in metaxy.toml
- [ ] All stores call `self.record_feature_graph_snapshot()` in `open()` when enabled
- [ ] Implement disk-based caching in `record_feature_graph_snapshot()` to prevent duplicate pushes
  - Use platformdirs to get appropriate cache directory
  - Cache file: `{cache_dir}/metaxy/push_cache.json`
  - Check cache before pushing to avoid duplicates in distributed setups
  - No expiry needed (snapshot_ids uniquely identify graph state)
- [ ] Default to false
- [ ] Add tests for auto-push and caching behavior

## Configuration
```toml
[core]
auto_push_graph = true
```

```bash
METAXY_AUTO_PUSH_GRAPH=1 python my_script.py
```

## Implementation Notes
- Modify MetadataStore.open() to check setting and call record_feature_graph_snapshot()
- Use `platformdirs.user_cache_dir("metaxy")` for cache location
- Cache format: `{snapshot_version: timestamp}` 
- If snapshot_version in cache, skip push (already recorded)

geoHeil:
resolves #56

### DIFF
--- a/src/metaxy/metadata_store/base.py
+++ b/src/metaxy/metadata_store/base.py
@@ -796,8 +796,6 @@ class MetadataStore(ABC):
         # Check if this exact snapshot already exists for this project
         snapshot_already_exists = False
         existing_spec_versions: dict[str, str] = {}
-        has_spec_version = False
-
         if existing_versions is not None:
             # Check if project column exists (it may not in old tables)
             if "project" in existing_versions.columns:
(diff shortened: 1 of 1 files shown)


geoHeil:
https://github.com/anam-org/metaxy/pull/225 should fix it

danielgafni (changes requested):
After thinking about it again, I don't think it's a good idea anymore. 

1. This should not be done when opening the store, because we would also be opening stores from other environments (e.g. prod from branch) to read from 

2. In general, since we cannot really use transactions, this functionality would create duplicates in system tables when running distributed processing. 

Let's close as not planned? Unless you know how to emulate a transaction reliably (I don't think it's possible? That's typically a property of the storage)

PR #231, opened 2025-11-05
Remove redundant tests

### LINKED ISSUE #139: [tests] Remove redundant tests
Currently a lot of tests are not needed: for example, a lot of APIs that should be agnostic to metadata store type are tested with fixtures parametrized over the store type. 

This slows down tests **by a lot**.  

We should remove such fixtures, maybe parametrize just over in-memory and one persistent store (duckdb). 

We have separate tests parametrize over stores types.

geoHeil:
resolves #139 

consider this a first version. There may be additional tests we want to remove.

Here the idea was to simplify this task by just reducing backend implementations - maybe this is not enough and we really do want to clean up the test cases.

### DIFF
--- a/tests/metadata_stores/test_resolve_update.py
+++ b/tests/metadata_stores/test_resolve_update.py
@@ -42,52 +42,12 @@ from metaxy.metadata_store import (
 from metaxy.metadata_store.clickhouse import ClickHouseMetadataStore
 from metaxy.metadata_store.duckdb import DuckDBMetadataStore
 from metaxy.models.plan import FeaturePlan
-from metaxy.provenance.types import HashAlgorithm, LazyIncrement
+from metaxy.provenance.types import HashAlgorithm
 
 # Type alias for feature plan output
 FeaturePlanOutput = tuple[
     FeatureGraph, Mapping[FeatureKey, type[BaseFeature]], FeaturePlan
 ]
-
-
-def verify_lazy_sql_frames(store: MetadataStore, lazy_result: LazyIncrement):
-    """Verify that lazy frames from IbisMetadataStore contain SQL and haven't been materialized.
-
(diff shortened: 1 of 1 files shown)


tests/metadata_stores/test_resolve_update.py
```diff
@@ -38,15 +42,339 @@
...
+from metaxy.provenance.types import HashAlgorithm, LazyIncrement
 
 # Type alias for feature plan output
 FeaturePlanOutput = tuple[
     FeatureGraph, Mapping[FeatureKey, type[BaseFeature]], FeaturePlan
 ]
 
 
-# ============= FEATURE GRAPH CONFIGURATIONS =============
+def verify_lazy_sql_frames(store: MetadataStore, lazy_result: LazyIncrement):
```
danielgafni: I don't think you need to add this: we already make an assertion with `df.implementation` (either in this test or golden provenance tests); but I will take a closer look

tests/metadata_stores/conftest.py
```diff
@@ -260,21 +260,41 @@ def case_clickhouse(
         return (ClickHouseMetadataStore, {"connection_string": clickhouse_db})
 
 
+class BasicStoreCases:
```
danielgafni: Yeah I like this change

tests/metadata_stores/test_resolve_update.py
geoHeil: removed

geoHeil:
I removed the thing from your comment - (i.e. addressed all open comments) but you wrote: Looking good so I thought you are fine with merging it.

Sorry - if this was too fast. Should we revert it?

danielgafni:
Yeah I think so

PR #234, opened 2025-11-05
postgres store

### LINKED ISSUE #223: [integrations] Add PostgreSQL support

geoHeil:
draft resolves #223 

### hashing

- PG does not support XXhash out of the box
- there are some addons to enable that
- but these are not well maintained
- therefore we use MD5 + sha256 (with pg_crypto) addon

### PG client libraries

Psycopg3 comes in 2 flavours - pure python and with a native C extension.
It looks like this difference is causing problems for IBIS as instead of strings - binary results are returned and IBIS fails to operate. Interestingly, this difference only seems to be a ting for Linux - not so for osx

For details, and an mini reproducible example see https://gist.github.com/geoHeil/577898303845428f54b904dad0cb0229

For now we have our own compatibility ship/monkeypatch which can translate.

## pulling out pieces

- bug fix https://github.com/anam-org/metaxy/pull/426
- cleanup deletion https://github.com/anam-org/metaxy/pull/427

### DIFF
--- a/src/metaxy/versioning/engine.py
+++ b/src/metaxy/versioning/engine.py
@@ -574,7 +574,7 @@ class VersioningEngine(ABC):
         This method can be overridden by subclasses (e.g., flat engines) to handle
         different column representations.
         """
-        return df.with_columns(  # ty: ignore[invalid-argument-type,invalid-return-type]
+        return df.with_columns(  # ty: ignore[invalid-argument-type]
             nw.col(METAXY_PROVENANCE).alias(METAXY_DATA_VERSION),
             nw.col(METAXY_PROVENANCE_BY_FIELD).alias(METAXY_DATA_VERSION_BY_FIELD),
         )
--- a/src/metaxy/versioning/flat_engine.py
+++ b/src/metaxy/versioning/flat_engine.py
@@ -84,7 +84,7 @@ class FlatVersioningMixin:
         )
 
         # For flattened columns, create corresponding data_version_by_field columns
-        current_columns = df.collect_schema().names()  # ty: ignore[invalid-argument-type]
+        current_columns = df.collect_schema().names()
         prov_prefix = f"{METAXY_PROVENANCE_BY_FIELD}__"
         data_prefix = f"{METAXY_DATA_VERSION_BY_FIELD}__"
 
@@ -93,9 +93,9 @@ class FlatVersioningMixin:
                 field_name = col.split("__", 1)[1]
                 target_col = f"{data_prefix}{field_name}"
                 if target_col not in current_columns:
-                    df = df.with_columns(nw.col(col).alias(target_col))  # ty: ignore[invalid-argument-type]
+                    df = df.with_columns(nw.col(col).alias(target_col))
 
-        return df  # ty: ignore[invalid-return-type]
+        return df
 
 
 class FlatVersioningEngine(FlatVersioningMixin, VersioningEngine, ABC):


src/metaxy/metadata_store/postgres.py
```diff
@@ -0,0 +1,152 @@
...
+
+if TYPE_CHECKING:
+    from metaxy.data_versioning.calculators.ibis import HashSQLGenerator
+    from metaxy.metadata_store.base import MetadataStore
+
+from metaxy.data_versioning.hash_algorithms import HashAlgorithm
+from metaxy.metadata_store.ibis import IbisMetadataStore
+
+
+class PostgresMetadataStore(IbisMetadataStore):
```
geoHeil: TODO add docs + example once we agree on the basics

src/metaxy/metadata_store/postgres.py
```diff
@@ -0,0 +1,257 @@
...
+        *,
+        host: str | None = None,
+        port: int | None = None,
+        user: str | None = None,
+        password: str | None = None,
+        database: str | None = None,
+        schema: str | None = None,
+        connection_params: dict[str, Any] | None = None,
+        fallback_stores: list["MetadataStore"] | None = None,
+        enable_pgcrypto: bool = True,
```
danielgafni: Default should be False: we should not attempt to silently manage infrastructure 

We might want to have the same flag with DuckDB by the way, but maybe there it isn't that critical

src/metaxy/metadata_store/postgres.py
```diff
@@ -0,0 +1,257 @@
...
+        super().open()
+
+        # Auto-enable pgcrypto if SHA256 is configured
+        if self.enable_pgcrypto and self.hash_algorithm == HashAlgorithm.SHA256:
+            self._ensure_pgcrypto_extension()
+
+    def _ensure_pgcrypto_extension(self) -> None:
+        """Ensure pgcrypto extension is enabled for SHA256 support.
+
+        Attempts to create the extension if it doesn't exist. Logs a warning
```
danielgafni: I think that's ok, but we need to clarify that automatic installation is not for production

tests/metadata_stores/conftest.py
```diff
@@ -16,6 +20,21 @@
 assert HashAlgorithmCases is not None  # ensure the import is not removed
 
 
+# Configure pytest-postgresql to find pg_ctl without using pg_config
+# (which can fail in Nix environments). Use shutil.which to find pg_ctl in PATH.
```
danielgafni: What is this referring to? 
pg_config doesn't fail in nix environments, it's just likely it's not set up correctly. Ping me, I'll give you an example

tests/metadata_stores/conftest.py
```diff
@@ -188,6 +195,64 @@ def clickhouse_db(clickhouse_server):
...
+def postgres_server(postgresql_proc: Any):
+    """Expose connection details for the pytest-postgresql server."""
+    try:
+        import ibis.backends.postgres  # noqa: F401
+    except ImportError:
+        pytest.skip("ibis-postgres not installed")
+
+    try:
+        import psycopg  # type: ignore[import]
+    except ImportError:
```
danielgafni: Same

tests/metadata_stores/conftest.py
```diff
@@ -188,6 +195,64 @@ def clickhouse_db(clickhouse_server):
...
+
+    test_conn_string = f"postgresql://{user}@{host}:{port}/{db_name}"
+
+    yield test_conn_string
+
+    try:
+        with psycopg.connect(admin_dsn, autocommit=True) as conn:
+            with conn.cursor() as cur:
+                cur.execute(f'DROP DATABASE IF EXISTS "{db_name}"')
+    except Exception:
```
danielgafni: Why silent exception handling?

tests/metadata_stores/test_postgres.py
```diff
@@ -0,0 +1,128 @@
...
+
+from typing import cast
+
+import pytest
+
+pytest.importorskip("ibis")
+
+try:
+    import ibis.backends.postgres  # noqa: F401
+except ImportError:
```
danielgafni: Remove

danielgafni (changes requested):
I think this looks mostly good! 

A few minor comments. 

Please avoid the silent exception handling inserted by LLMs all the time. As well as catching the entire Exception class. It's their way to get things "working" when actually they may be not behaving as expected. 

And let's resolve the Nix thing, I can send you a working config.

src/metaxy/metadata_store/postgres.py
geoHeil: done; for duckdb - I think we do not necessarily need that flag as there it is easier to install the addons/more common. However, in the case of an airgapped installation things would indeed get tricky as there is not yet any (at least not to me known) artifactory for proxying the duckdb addon ecosystem

src/metaxy/metadata_store/postgres.py
```diff
@@ -0,0 +1,257 @@
...
+                hash_selects.append(f"{hash_expr} as {hash_col}")
+
+            hash_clause = ", ".join(hash_selects)
+            table_sql = table.compile()
+            return f"SELECT *, {hash_clause} FROM ({table_sql}) AS __metaxy_temp"
+
+        generators[HashAlgorithm.SHA256] = sha256_generator
+        return generators
+
+    def display(self) -> str:
```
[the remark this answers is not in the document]
geoHeil: Normally for IBIS there is a) the python typed IBIS DSL or b) the raw string where it is on us to ensure that no SQL injection took place

src/metaxy/metadata_store/postgres.py
```diff
@@ -0,0 +1,257 @@
...
+            **kwargs: Passed to [metaxy.metadata_store.ibis.IbisMetadataStore][]
+                (e.g., `hash_algorithm`, `auto_create_tables`).
+
+        Raises:
+            ValueError: If neither connection_string nor connection parameters provided.
+            ImportError: If Ibis or psycopg2 driver not installed.
+
+        Note:
+            When using SHA256 hash algorithm, pgcrypto extension is required.
+            The store attempts to enable it automatically on open unless
```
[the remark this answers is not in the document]
geoHeil: addressed

src/metaxy/metadata_store/postgres.py
```diff
@@ -0,0 +1,257 @@
...
+        """
+        try:
+            # Use underlying connection to execute DDL (Ibis doesn't expose CREATE EXTENSION)
+            # For PostgreSQL backend, conn.con is the psycopg2 connection
+            raw_conn = self.conn.con  # pyright: ignore[reportAttributeAccessIssue]
+            with raw_conn.cursor() as cursor:  # pyright: ignore[reportAttributeAccessIssue]
+                cursor.execute("CREATE EXTENSION IF NOT EXISTS pgcrypto")
+                raw_conn.commit()  # pyright: ignore[reportAttributeAccessIssue]
+            logger.debug("pgcrypto extension enabled successfully")
+        except Exception as e:
```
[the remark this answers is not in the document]
geoHeil: addressed

tests/metadata_stores/conftest.py
geoHeil: dropped

tests/metadata_stores/test_postgres.py
geoHeil: dropped

src/metaxy/metadata_store/postgres.py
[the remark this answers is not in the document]
geoHeil: but we cannot directly pull it into init as we do not yet have opened the PG connection there;

further I would imagine if the init is heavy that metaxy addons like dagster would possibly be slow during init

tests/metadata_stores/conftest.py
geoHeil: but for me pg_config seems to still point to my local (non-nix brew PG installation); so I would (at least with the current flake) guess that the extra steps are needed

.github/workflows/QA.yml
```diff
@@ -11,6 +11,12 @@ on:
       - dev
   pull_request:
   workflow_dispatch:
+    inputs:
+      debug_enabled:
+        type: boolean
```
geoHeil: should the leave this in? it is actually really useful for debugging - but if we leave it in you do have to disable that forked repos can trigger CI without review from the core team. Otherwise it is a security issue

tests/metadata_stores/conftest.py
```diff
@@ -15,6 +26,42 @@
...
+
+def _find_free_port() -> int:
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+        s.bind(("127.0.0.1", 0))
+        s.listen(1)
+        port = s.getsockname()[1]
+    return port
+
+
+def _format_pg_connection_string(
```
danielgafni: Curious if there is one available in psycopg2 or ibis, but I guess it's not very important

danielgafni (commented):
Great job with the flake! 
Happy it's all working now. Thanks for cleaning it up properly. 

I left minor comments. 

A question about tests: why do we have to define so many PostgreSQL fixtures? pytest-postgresql provides them out of the box. 

Do they not work in our setup?

src/metaxy/metadata_store/ibis.py
```diff
@@ -178,6 +186,32 @@ def _create_provenance_tracker(self, plan):
...
             pass
 
+    @staticmethod
+    def _ensure_string_identifier(value: str | bytes) -> str:
+        """Ensure SQL identifier inputs are plain strings."""
+        if isinstance(value, bytes):
+            return value.hex()
+        return str(value)
+
+    def _get_hash_sql_generators(self) -> dict[HashAlgorithm, HashSQLGenerator]:
```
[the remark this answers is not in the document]
geoHeil: fixed

tests/metadata_stores/conftest.py
geoHeil: removed

src/metaxy/provenance/ibis.py
```diff
@@ -45,6 +50,18 @@ def __init__(
         super().__init__(plan)
         self.hash_functions: dict[HashAlgorithm, IbisHashFn] = hash_functions
 
+    @staticmethod
+    def _ensure_ibis_table(df: FrameT) -> ibis.expr.types.Table:
+        """Return the underlying Ibis table, ensuring the frame is Ibis-backed."""
+        import ibis.expr.types
+
+        native = df.to_native()
+        if not isinstance(native, ibis.expr.types.Table):
```
danielgafni: I'd check df.implementation before the .to_native() call instead. It's more lightweight.

src/metaxy/migrations/detector.py
```diff
@@ -155,6 +155,8 @@ def detect_migration(
 
     yaml_path = migrations_dir / f"{migration_id}.yaml"
     migration_yaml = {
+        "migration_type": migration.migration_type,  # Required for discriminated union
```
danielgafni: Looks like irrelevant changes?

src/metaxy/metadata_store/postgres.py
```diff
@@ -0,0 +1,972 @@
...
+# installed instead of the C-optimized version. The pure-Python version has a
+# different default behavior: it returns `bytes` for certain metadata query
+# results (like schema names and data types), whereas the C version correctly
+# decodes them to `str`. This inconsistency breaks upstream libraries like Ibis.
+#
+# SOLUTION: This code block acts as a compatibility shim. It monkey-patches the
+# `psycopg.Cursor` class to normalize the data flow. It ensures that both
+# query parameters (on the way in) and query results (on the way out) are
+# consistently strings, regardless of which `psycopg` implementation is in use.
+# This patch is applied globally and exactly once when this module is imported.
```
danielgafni: I don't feel good about patching something globally. It will definitely break things for users.

src/metaxy/metadata_store/postgres.py
```diff
@@ -0,0 +1,972 @@
...
+    from metaxy.models.types import FeatureKey
+
+
+_PGCRYPTO_ERROR_TYPES: tuple[type[Exception], ...] = (_PsycopgError,)
+SchemaMapping = Mapping[str, PolarsDataType | PolarsDataTypeClass]
+
+logger = logging.getLogger(__name__)
+
+try:
+    if not getattr(PsycopgCursor, "_metaxy_compat_patched", False):
```
danielgafni: I still need to dig into this issue, but I definitely don't feel good about just magically patching a global object to act differently. 

If this fix is truly needed, I can see 2 ways we could apply it: 
- scope it only to our metadata store. We should not affect other libraries and user code. 
- if that's not possible, at least make it opt-in via a configurable parameter somewhere

src/metaxy/metadata_store/postgres.py
```diff
@@ -0,0 +1,972 @@
...
+                self._pgcrypto_extension_checked = False
+                import sys
+
+                if not self._has_native_struct_support():
+                    self._struct_compat_mode = True
+                    message = (
+                        "!!! Metaxy WARNING: PostgreSQL backend lacks native STRUCT type support. "
+                        "Falling back to JSON serialization compatibility mode. !!!"
+                    )
+                    print(message, file=sys.stderr)
```
danielgafni: What is that for? We already have the logger.warning below

danielgafni (changes requested):
I didn't have enough time to do a full review but here are some comments. 

By the way, I'm skeptical about the shim thing. 

If that's a bug in other libraries it shouldn't be our job to fix it. 

Can we check if psycopg2 works without this shim? Is it only psycopg3's issue? I would rather not support it yet then.

src/metaxy/metadata_store/postgres.py
geoHeil: removed

src/metaxy/metadata_store/base.py
```diff
@@ -185,12 +185,12 @@ def resolve_update(
         Args:
             feature: Feature class to resolve updates for
             samples: Pre-computed DataFrame with ID columns
-                and `PROVENANCE_BY_FIELD_COL` column. When provided, `MetadataStore` skips upstream loading, joining,
+                and `[[metaxy.models.constants.]]` column. When provided, `MetadataStore` skips upstream loading, joining,
```
geoHeil: TODO fix this

I have a fix locally but cannot push

```
 git push
ERROR: user:1694964:geoHeil
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

Anyways:

```
Must have `[[metaxy.models.constants.METAXY_PROVENANCE_BY_FIELD]]`
```

is not correctly interpreted from mkdocs

docs/integrations/index.md
```diff
@@ -55,6 +55,18 @@ Learn more about metadata stores [here](../learn/metadata-stores.md).
 
     [:octicons-arrow-right-24: API docs](../reference/api/metadata-stores/ibis/clickhouse.md)
 
+-   :custom-postgresql:{ .lg .middle } [**PostgreSQL**](metadata-stores/databases/postgres.md)
+
+    ---
+
+    :material-tag-outline: Database
+
+    Use PostgreSQL as a transactional metadata store with strong consistency guarantees.
```
danielgafni: I think this is strange to highlight here since Metaxy is not using any transactions (except just inserting data).   
  
I'd rather word it just as "Use PostgreSQL to store Metaxy metadata."

src/metaxy/metadata_store/postgres.py
```diff
@@ -0,0 +1,1158 @@
...
+        # Call parent context manager to establish connection
+        with super().open(mode):
+            try:
+                # Reset pgcrypto check for the new connection
+                self._pgcrypto_extension_checked = False
+
+                if not self._has_native_struct_support():
+                    self._struct_compat_mode = True
+                    message = (
+                        "!!! Metaxy WARNING: PostgreSQL backend lacks native STRUCT type support. "
```
danielgafni: should use `warnings.warn`​, see how other warnings are emitted

src/metaxy/metadata_store/postgres.py
```diff
@@ -0,0 +1,1158 @@
...
+    def _has_native_struct_support(self) -> bool:
+        """
+        NOTE: Native struct support for PostgreSQL in the Ibis/sqlglot stack has proven
+        unreliable across different environments. We are temporarily disabling it
+        to enforce the more stable JSON serialization path.
+        """
+        return False
+
+    @staticmethod
+    def _serialize_struct_column(df: pl.DataFrame, column_name: str) -> pl.DataFrame:
```
danielgafni: Looks like this method should belong somewhere upstream, it's not specific to Postgresql.

PR #235, opened 2025-11-05
[core] add DeltaLakeMetadataStore

### LINKED ISSUE #224: [store] deltalake

geoHeil:
resolves #224

### DIFF
--- a/tests/metadata_stores/conftest.py
+++ b/tests/metadata_stores/conftest.py
@@ -30,7 +30,7 @@ assert HashAlgorithmCases is not None  # ensure the import is not removed
 def find_free_port() -> int:
     """Find a free port on the system."""
     with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
-        s.bind(("", 0))
+        s.bind(("127.0.0.1", 0))
         s.listen(1)
         port = s.getsockname()[1]
     return port


src/metaxy/metadata_store/delta.py
```diff
@@ -0,0 +1,304 @@
...
+        if not table_exists and not self.auto_create_tables:
+            raise TableNotFoundError(
+                f"Delta table does not exist for feature {feature_key.to_string()} at {table_path}. "
+                f"Enable auto_create_tables=True to automatically create tables."
+            )
+
+        # Create parent directory if needed
+        table_path.parent.mkdir(parents=True, exist_ok=True)
+
+        arrow_table = df.to_arrow()
```
danielgafni: I think we should be handling Polars natively. 

In the future I want "write_metadata" method to take Narwhals frames including lazy polars frames. We should sink them.

src/metaxy/metadata_store/delta.py
```diff
@@ -0,0 +1,304 @@
...
+
+        # Check if table exists
+        if not table_exists and not self.auto_create_tables:
+            raise TableNotFoundError(
+                f"Delta table does not exist for feature {feature_key.to_string()} at {table_path}. "
+                f"Enable auto_create_tables=True to automatically create tables."
+            )
+
+        # Create parent directory if needed
+        table_path.parent.mkdir(parents=True, exist_ok=True)
```
danielgafni: Won't work with object stores, use deltalake native APIs instead

mkdocs.yml
```diff
@@ -189,13 +189,14 @@ nav:
...
-        - DuckDB: reference/api/metadata-stores/ibis/duckdb.md
-        - ClickHouse: reference/api/metadata-stores/ibis/clickhouse.md
-      - Memory: reference/api/metadata-stores/memory.md
+      - Metadata Stores:
+        - reference/api/metadata-stores/index.md
+        - Databases:
+          - reference/api/metadata-stores/ibis/index.md
+          - DuckDB: reference/api/metadata-stores/ibis/duckdb.md
+          - ClickHouse: reference/api/metadata-stores/ibis/clickhouse.md
+        - Delta Lake: reference/api/metadata-stores/delta.md
```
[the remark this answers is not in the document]
geoHeil: Arent they already in the section of Metadata stores?

src/metaxy/metadata_store/delta.py
```diff
@@ -0,0 +1,304 @@
...
+
+    _should_warn_auto_create_tables = False
+
+    def __init__(
+        self,
+        root_path: str | Path,
+        *,
+        storage_options: dict[str, Any] | None = None,
+        fallback_stores: list[MetadataStore] | None = None,
+        **kwargs: Any,
```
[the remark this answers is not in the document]
geoHeil: we have now object_store_kwargs

src/metaxy/metadata_store/delta.py
```diff
@@ -0,0 +1,304 @@
...
+        return FeatureKey(parts)
+
+    def _feature_path(self, feature_key: FeatureKey) -> Path:
+        """Get the filesystem path for a feature's Delta table."""
+        table_name = feature_key.table_name
+        return self.root_path / table_name
+
+    def _table_exists(self, table_path: Path) -> bool:
+        """Check if a Delta table exists at the given path."""
+        return (table_path / "_delta_log").exists()
```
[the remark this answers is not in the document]
geoHeil: should be fixed now

src/metaxy/metadata_store/delta.py
```diff
@@ -0,0 +1,304 @@
...
+                mode="append",
+                schema_mode="merge",
+                storage_options=self.storage_options if self.storage_options else None,
+            )
+        except Exception as e:
+            raise RuntimeError(
+                f"Failed to write metadata for feature {feature_key.to_string()}: {e}"
+            ) from e
+
+    def _drop_feature_metadata_impl(self, feature_key: FeatureKey) -> None:
```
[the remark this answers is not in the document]
geoHeil: fixed now

pyproject.toml
```diff
@@ -43,6 +43,10 @@ clickhouse = [
 duckdb = [
   "ibis-framework[duckdb]>=11.0.0",
 ]
+delta = [
+  "deltalake>=1.2.1",
+  "obstore>=0.8.2",
```
danielgafni: obstore shouldn't be a dependency, deltalake works without it

danielgafni (changes requested):
Deltalake doesn't need additional dependencies to work. Use its native APIs. 

https://delta-io.github.io/delta-rs/api/delta_table/#deltalake.DeltaTable

For example, you can use table.is_deltatable() to check for existence or table.create() to create one. 

Do not attempt to create or delete files or directories manually for the users. But if a native API has something like mkdir=True - it should be ok to use it.

src/metaxy/metadata_store/delta.py
geoHeil: see the current approach - it verifies connectivity

pyproject.toml
geoHeil: not even for the deletion of full tables/features?

danielgafni:
Also, explore without_files=True for reading or appending data 

---

If True, will load table without tracking files. Some append-only applications might have no need of tracking any files. So, the DeltaTable will be loaded with a significant memory reduction.

src/metaxy/metadata_store/base.py
```diff
@@ -497,6 +500,12 @@ def write_metadata(
...
 
+        # Sink lazy frames to eager DataFrames
+        if isinstance(df, nw.LazyFrame):
+            df = df.collect()
+        if isinstance(df, pl.LazyFrame):
+            df = df.collect()
+
         # Convert Narwhals to Polars if needed
         if isinstance(df, nw.DataFrame):
             df = df.to_polars()
```
danielgafni: <p class='graphite__hidden'><i>[Re: line +523]</i></p>
Side note: I feel like this logic is backwards, we need to do it the other way. 
Metaxy internals should not be in the way of a normal user code path. 
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/anam-org/metaxy/235?utm_source=unchanged-line-comment">Graphite</a>.</p>

src/metaxy/metadata_store/delta.py
```diff
@@ -0,0 +1,447 @@
...
+                    if parsed.netloc and parsed.netloc not in {"", "localhost"}:
+                        local_path_str = f"//{parsed.netloc}{local_path_str}"
+                    local_path = Path(local_path_str).expanduser()
+                    self._local_root_path = local_path
+                    self._root_uri = str(local_path)
+                else:
+                    self._local_root_path = None
+                    sanitized = root_str.rstrip("/")
+                    self._root_uri = sanitized
+                    self._is_remote = True
```
danielgafni: Can you explain what's about all this logic? that's really **a lot** of string/path manipulation with unclear purpose. 

A DeltaLake table just needs a single URI to be created. It means we only need to store the root path string and `os.path.join` it with the feature key to get a feature-specific table. This should be 1 line in `__init__` and 1 line in `get_table_path` or similar method.

src/metaxy/metadata_store/delta.py
```diff
@@ -0,0 +1,447 @@
...
+    ]:
+        """Delta Lake store does not provide native SQL execution."""
+        raise NotImplementedError(
+            "DeltaMetadataStore does not support native field provenance calculations"
+        )
+
+    def open(self) -> None:
+        """Ensure root directory exists and create system tables if needed."""
+        if self._local_root_path is not None:
+            self._local_root_path.mkdir(parents=True, exist_ok=True)
```
danielgafni: bad call

src/metaxy/metadata_store/delta.py
```diff
@@ -0,0 +1,447 @@
...
+            self._create_system_tables()
+
+    def close(self) -> None:
+        """No persistent resources to release."""
+        # delta-rs is used in one-shot write/read calls, so nothing to close.
+        pass
+
+    # ===== Internal helpers =====
+
+    def _create_system_tables(self) -> None:
```
danielgafni: Again, DeltaLake auto-creates tables anyway, I don't think we need this logic here

src/metaxy/metadata_store/delta.py
```diff
@@ -0,0 +1,447 @@
...
+            storage_options=self._storage_options_payload(),
+        )
+
+        # Use column projection for efficiency if columns are specified
+        if columns is not None:
+            # Need to ensure system columns are included for filtering
+            cols_to_read = set(columns)
+            if feature_version is not None:
+                cols_to_read.add("metaxy_feature_version")
+            arrow_table = delta_table.to_pyarrow_table(columns=list(cols_to_read))
```
danielgafni: This means all the reads are going to be eager. 
We have to use `pl.scan_delta` instead to enable lazy reads.

tests/metadata_stores/test_delta.py
```diff
@@ -0,0 +1,165 @@
+"""Delta Lake-specific tests that don't apply to other stores."""
+
+from __future__ import annotations
+
+import polars as pl
+import pytest
+
+pytest.importorskip("pyarrow")
```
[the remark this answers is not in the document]
danielgafni: just drop the importstkip, right now we do not have optional tests

tests/metadata_stores/test_delta.py
```diff
@@ -0,0 +1,257 @@
...
+pytest.importorskip("deltalake")
+
+from deltalake import DeltaTable
+
+from metaxy._utils import collect_to_polars
+from metaxy.metadata_store.delta import DeltaMetadataStore
+from metaxy.models.types import FeatureKey
+
+
+def _delta_log_versions(path) -> set[str]:
```
danielgafni: why is that needed, can't we use native APIs instead?

tests/metadata_stores/test_provenance_golden_reference.py
```diff
@@ -229,6 +230,20 @@ def case_clickhouse(self, hash_algorithm: HashAlgorithm, clickhouse_db: str):
                 f"Hash algorithm {hash_algorithm} not supported by {ClickHouseMetadataStore}"
             )
 
+    @parametrize_with_cases("hash_algorithm", cases=HashAlgorithmCases)
+    def case_delta(self, hash_algorithm: HashAlgorithm, tmp_path):
```
danielgafni: I think this metadata store should not be included in this test. It is not involved with provenance calculation. 

It should only be tested for metadata operations, e.g. read/write/delete. 

We should have a similar, standard test suite for this operations, and we should run them through all our metadata stores. 

This standard test suite should be created in another PR.

danielgafni (changes requested):
We still need moto tests - it will give you an S3-compatible local `AWS_ENDPOINT_URL` and other connections params that would work as with a real S3 bucket.

src/metaxy/metadata_store/base.py
```diff
@@ -480,6 +480,12 @@ def write_metadata(
         if not is_system_table:
             self._validate_project_write(feature)
 
+        # Sink lazy frames to eager DataFrames
+        if isinstance(df, nw.LazyFrame):
```
geoHeil: cant we do this lazy?

mkdocs.yml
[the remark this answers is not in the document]
geoHeil: `Stores without engine`?

.pre-commit-config.yaml
```diff
@@ -11,6 +11,7 @@ repos:
       - id: check-yaml
       - id: check-added-large-files
         exclude: ^.basedpyright/baseline\.json$
+        args: ['--maxkb=600']
```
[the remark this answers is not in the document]
geoHeil: done

src/metaxy/metadata_store/delta.py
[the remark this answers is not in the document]
geoHeil: simplified now

src/metaxy/metadata_store/base.py
```diff
@@ -480,18 +489,65 @@ def write_metadata(
...
+            df = df.to_native()  # type: ignore[assignment]
+
+        if pa is not None and isinstance(df, pa.Table):
+            # Streaming paths may yield Arrow tables; normalize to Polars.
+            df = cast(pl.DataFrame, pl.from_arrow(df))
+
+        # Now df is either pl.DataFrame or pl.LazyFrame
+
+        # Collect lazy frames unless subclass wants to handle them
+        if self._auto_collect_lazy_frames and isinstance(df, pl.LazyFrame):
```
danielgafni: I don't really like this pattern in the base store. 

We **never** collect DataFrames in the whole base store code path. There is only one exception, and it emits a warning, and we make sure it never happens in tests unless explicitly slowed: when falling back to the Polars provenance tracker from the native one. 

There is basically no reason to perform this collect, am I wrong here? Even if we decide to introduce this option, I would default it to False - that's just safer.

tests/metadata_stores/test_delta.py
```diff
@@ -0,0 +1,257 @@
...
+                "sample_uid": [1, 2, 3],
+                "metaxy_provenance_by_field": [
+                    {"frames": "h1", "audio": "h1"},
+                    {"frames": "h2", "audio": "h2"},
+                    {"frames": "h3", "audio": "h3"},
+                ],
+            }
+        ).lazy()
+
+        with store.allow_cross_project_writes():
```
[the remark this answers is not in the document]
geoHeil: removed

docs/reference/api/metadata-stores/delta.md
```diff
@@ -0,0 +1,95 @@
+---
+title: Delta Lake Metadata Store
+---
+
+# DeltaMetadataStore
+
+::: metaxy.metadata_store.delta.DeltaMetadataStore
+
+## Installation
```
danielgafni: Should come before the API reference OR be in `learn` docs instead

docs/reference/api/metadata-stores/delta.md
```diff
@@ -0,0 +1,95 @@
...
+            "secret_access_key": "...",
+            "region": "us-west-2",
+        }
+    },
+)
+```
+
+`storage_options` go straight to `deltalake`.
+
+### Storage Layout
```
danielgafni: This is something I would like to have more control over: let's introduce a `layout` option with two possible values: `flat` and `nested`. 

`flat` would be the current implementation. 
`nested` would store deltalake tables `in/feature/key/directories`.

Should only affect how `_feature_uri` is implemented.

src/metaxy/metadata_store/base.py
```diff
@@ -7,12 +7,17 @@
...
 from contextlib import contextmanager
 from datetime import datetime, timezone
-from typing import TYPE_CHECKING, Any, Literal, overload
+from typing import Any, Literal, overload
 
 import narwhals as nw
 import polars as pl
 from typing_extensions import Self
 
+try:  # pyarrow is an optional dependency needed only for Arrow streaming paths
```
danielgafni: can this be moved from the top-level imports to the specific part of the code that's actually using it?

src/metaxy/metadata_store/base.py
```diff
@@ -480,30 +483,65 @@ def write_metadata(
...
+        if not isinstance(df, (pl.DataFrame, pl.LazyFrame)):
+            raise TypeError(
+                f"Expected Polars DataFrame or LazyFrame after conversion, got {type(df)}"
+            )
 
-        # For system tables, write directly without feature_version tracking
+        # For system tables, collect lazy frames and write directly
         if is_system_table:
+            if isinstance(df, pl.LazyFrame):
+                df = df.collect()
```
danielgafni: I'm not sure, why exactly is this placed here? shouldn't `_write_metadata_impl` handle collecting if it's really needed? Collection should be deferred as long as possible

src/metaxy/metadata_store/base.py
```diff
@@ -515,34 +553,73 @@ def write_metadata(
...
+        # Get snapshot_version from active graph
+        from metaxy.models.feature import FeatureGraph
+
+        graph = FeatureGraph.get_active()
+        current_snapshot_version = graph.snapshot_version
+
+        # Add system columns
+        return df.with_columns(
+            [
+                pl.lit(current_feature_version).alias(FEATURE_VERSION_COL),
```
danielgafni: should be `nw.lit` (see comment above)

src/metaxy/metadata_store/delta.py
```diff
@@ -0,0 +1,453 @@
...
+    """
+    Delta Lake metadata store backed by [delta-rs](https://github.com/delta-io/delta-rs).
+
+    Stores each feature's metadata in a dedicated Delta table located under ``root_path``.
+    Uses Polars/Narwhals components for metadata operations and relies on delta-rs for persistence.
+
+    **Lazy Frame Handling:**
+
+    - **Eager DataFrames**: Written directly using Polars' native write_delta (efficient)
+    - **Lazy Frames with streaming_chunk_size=None** (default): Collected once then written
```
danielgafni: items in the list have to be separated by newlines

src/metaxy/metadata_store/delta.py
```diff
@@ -0,0 +1,453 @@
...
+            table_uri,
+            storage_options=self.storage_options or None,
+        )
+
+    # ===== Storage operations =====
+
+    def _write_metadata_impl(
+        self,
+        feature_key: FeatureKey,
+        df: pl.DataFrame | pl.LazyFrame,
```
danielgafni: Can we also handle narwhals frames? just convert them to polars.

src/metaxy/metadata_store/delta.py
```diff
@@ -0,0 +1,453 @@
...
+
+        # Apply column selection early if specified (pushdown)
+        if cols_to_read is not None:
+            lf = lf.select(cols_to_read)
+
+        # Convert to Narwhals
+        nw_lazy = nw.from_native(lf)
+
+        # Apply filters
+        if feature_version is not None:
```
danielgafni: This has to be moved to the base store as other generic filters

danielgafni (changes requested):
A good chunk of comments :)

src/metaxy/metadata_store/base.py
geoHeil: only if we add pyarrow to default deps. should we do that? otherwise we would need this fallback

docs/reference/api/metadata-stores/delta.md
geoHeil: done

docs/reference/api/metadata-stores/delta.md
geoHeil: done

docs/reference/api/metadata-stores/delta.md
```diff
@@ -0,0 +1,95 @@
...
+The Delta Lake backend relies on [`deltalake`](https://delta-io.github.io/delta-rs/python/), which is provided through Metaxy’s `delta` dependency group.
+Install it with:
+
+```bash
+uv sync --group delta
+```
+
+This will pull in the delta-rs bindings alongside the core Metaxy dependencies (if they are not already installed).
+
+## Usage
```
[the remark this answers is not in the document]
geoHeil: moved

src/metaxy/metadata_store/base.py
geoHeil: fixed

src/metaxy/metadata_store/base.py
geoHeil: done

docs/reference/api/metadata-stores/delta.md
```diff
@@ -0,0 +1,42 @@
...
+Delta Lake is the recommended object-store-backed metadata backend.
+It keeps one Delta table per feature, appends new metadata versions, and leans on Polars / Narwhals for provenance calculations.
+
+## Installation
+
+The backend relies on [`deltalake`](https://delta-io.github.io/delta-rs/python/)
+which ships with Metaxy’s `delta` dependency group:
+
+```bash
+uv sync --group delta
```
danielgafni: This is obviously wrong for public docs, could you please proofread them?

docs/reference/api/metadata-stores/delta.md
```diff
@@ -0,0 +1,42 @@
...
+```
+
+## Object Stores
+
+Point `root_path` at any supported URI (`s3://`, `abfss://`, `gs://`, `file://`,
```
danielgafni: Do not specify them all here, instead link to DeltaLake docs (I'm pretty sure they in turn link to obstore)

docs/reference/api/metadata-stores/delta.md
```diff
@@ -0,0 +1,42 @@
...
+Control how feature keys map to Delta directories with `layout`:
+
+- `flat` (default) keeps the current `<namespace>__<feature>` directory scheme.
+- `nested` places every key part into its own directory (e.g.
+  `namespace/feature/channel`), which can make large hierarchies easier to
+  inspect in object-store consoles.
+
+Only URI generation changes; table contents remain identical.
+
+## Hashing
```
danielgafni: There is no need to repeat it here, it's already obvious from API docs

docs/reference/api/metadata-stores/delta.md
```diff
@@ -0,0 +1,42 @@
...
+
+`DeltaMetadataStore` defaults to `HashAlgorithm.XXHASH64`, in line with the other non-SQL stores.
+Use the constructor arguments to override the algorithm if you need a different hash.
+
+::: metaxy.metadata_store.delta.DeltaMetadataStore
```
danielgafni: Maybe insert a visual separator such as ``` before the API docs

src/metaxy/metadata_store/base.py
```diff
@@ -515,34 +567,77 @@ def write_metadata(
...
-            # Get snapshot_version from active graph
+        # Get current feature version
+        if isinstance(feature, type) and issubclass(feature, BaseFeature):
+            current_feature_version = feature.feature_version()  # type: ignore[attr-defined]
+        else:
             from metaxy.models.feature import FeatureGraph
 
             graph = FeatureGraph.get_active()
-            current_snapshot_version = graph.snapshot_version
+            feature_cls = graph.features_by_key[feature_key]
```
danielgafni: I am pretty sure there already is a helper method that does this conversion, let's reuse it

src/metaxy/metadata_store/base.py
```diff
@@ -583,11 +678,34 @@ def _validate_schema(self, df: pl.DataFrame) -> None:
...
+        columns: Sequence[str] | None,
+    ) -> nw.LazyFrame[Any]:
+        """Apply standard filters/column projections to lazy frames."""
+        if feature_version is not None:
+            lazy_frame = lazy_frame.filter(
+                nw.col(FEATURE_VERSION_COL) == feature_version
+            )
+
+        if filters is not None:
+            for expr in filters:
```
danielgafni: Can be replaced with 

```
df = df.filter(*exprs)
```

src/metaxy/metadata_store/base.py
```diff
@@ -583,11 +678,34 @@ def _validate_schema(self, df: pl.DataFrame) -> None:
                 f"DataFrame must have '{SNAPSHOT_VERSION_COL}' column"
             )
 
-    def _validate_schema_system_table(self, df: pl.DataFrame) -> None:
+    def _validate_schema_system_table(self, df: pl.DataFrame | pl.LazyFrame) -> None:
         """Validate schema for system tables (minimal validation)."""
         # System tables don't need metaxy_provenance_by_field column
         pass
 
+    def _apply_read_filters(
```
danielgafni: I'm not sure if this is needed anymore, upstream dependency filtering has been moved to `metaxy.provenance` class

src/metaxy/metadata_store/delta.py
```diff
@@ -0,0 +1,484 @@
...
+    delta-rs write call.
+
+    **Lazy Frame Handling:**
+
+    - **Eager DataFrames**: Written directly using Polars' native write_delta (efficient)
+
+    - **Lazy Frames with streaming_chunk_size=None** (default): Collected once then written
+
+    - **Lazy Frames with streaming_chunk_size=N**: Streamed in batches of size N without full collection,
+      enabling processing of datasets larger than RAM (at the cost of slower performance and more Delta versions)
```
danielgafni: "more delta versions" change to "creating multiple DeltaLake versions within a single sink operation" (or similar)

src/metaxy/metadata_store/delta.py
```diff
@@ -0,0 +1,484 @@
...
+
+    Example:
```
danielgafni: It probably makes sense to have 2 examples: 

```
Example: Providing Object Storage Credentials 
    
    ...


Example: Sinking Lazy Frames 

    ...
```

src/metaxy/metadata_store/delta.py
```diff
@@ -0,0 +1,484 @@
...
+
+        # Streaming mode for large datasets
+        store = DeltaMetadataStore(
+            "/data/metaxy/metadata",
+            streaming_chunk_size=10000,  # Stream in 10k row batches
+            storage_options={"AWS_REGION": "us-west-2"},
+        )
+
+        with store:
+            with store.allow_cross_project_writes():
```
danielgafni: This allow_cross_project_writes thing doesn't belong here

src/metaxy/metadata_store/delta.py
```diff
@@ -0,0 +1,484 @@
...
+
+                - ✗ More Delta versions (one per batch)
+
+                - ✗ Uses unstable Polars API
+
+                For most use cases, leave as None (default).
+            layout: Directory layout used when deriving table paths from feature keys.
+                - **"flat"**: Current `<namespace>__<feature>` directories (backward compatible default)
+
+                - **"nested"**: `namespace/feature/...` directories for easier navigation in consoles.
```
danielgafni: Remove "for easier navigation in consoles"

src/metaxy/metadata_store/delta.py
```diff
@@ -0,0 +1,484 @@
...
+
+                - ✗ Slower performance (Polars sink_batches overhead)
+
+                - ✗ More Delta versions (one per batch)
+
+                - ✗ Uses unstable Polars API
+
+                For most use cases, leave as None (default).
+            layout: Directory layout used when deriving table paths from feature keys.
+                - **"flat"**: Current `<namespace>__<feature>` directories (backward compatible default)
```
danielgafni: Backwards compatible with what?

src/metaxy/metadata_store/delta.py
```diff
@@ -0,0 +1,484 @@
...
+
+    def _feature_uri(self, feature_key: FeatureKey) -> str:
+        """Return the URI/path used by deltalake for this feature."""
+        relative = self._relative_feature_path(feature_key)
+        if self._local_root_path is not None:
+            return str(self._local_root_path / Path(relative))
+        base = self._root_uri.rstrip("/")
+        return f"{base}/{relative}".rstrip("/")
+
+    def _feature_local_path(self, feature_key: FeatureKey) -> Path | None:
```
danielgafni: Should not be a thing

src/metaxy/metadata_store/delta.py
```diff
@@ -0,0 +1,484 @@
...
+        ):
+            return
+
+        # Load the Delta table
+        delta_table = self._get_delta_table(feature_key, include_files=False)
+
+        # Use Delta's delete operation - soft delete all rows
+        # This marks rows as deleted in transaction log without physically removing files
+        delta_table.delete()
+        self._invalidate_delta_table_cache(feature_key)
```
danielgafni: Why should we be invalidating this cache here? The DeltaTable object is still valid.

danielgafni (changes requested):
Please proofread the docs. I found lots of small issues with them, mostly AIsms. I really don't want to be doing that. 

Also, maybe consider using `functools.lru_cache` instead of the hand-rolled dict? Would save us hundreds of LOC. 

Lastly, I believe the "search" functionality has to be dropped, we can't do it in the object store anyway. That's not something we should be doing. We should only work with known locations determined by the feature keys (known feature keys).

src/metaxy/metadata_store/delta.py
```diff
@@ -0,0 +1,484 @@
...
+                    f"Could not parse Delta directory '{feature_dir}' as FeatureKey: {exc}",
+                    UserWarning,
+                    stacklevel=2,
+                )
+                continue
+            if not self._is_system_table(feature_key):
+                feature_keys.append(feature_key)
+        return sorted(feature_keys)
+
+    def _list_features_from_object_store(self) -> list[FeatureKey]:
```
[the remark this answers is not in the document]
geoHeil: similar issue also there https://github.com/anam-org/metaxy/pull/218#issuecomment-3528271522

src/metaxy/metadata_store/base.py
geoHeil: see the modification only imported in the specific code path that needs it (fallback stores with diff resolution)

src/metaxy/metadata_store/base.py
```diff
@@ -480,30 +483,65 @@ def write_metadata(
...
+        self._validate_schema_for_write(df)
+
+        # Write metadata - delegate to store (store decides whether to collect lazy frames)
+        self._write_metadata_impl(feature_key, df)
+
+    def _add_system_columns(
+        self,
+        feature: FeatureKey | type[BaseFeature],
+        feature_key: FeatureKey,
+        df: pl.DataFrame | pl.LazyFrame,
```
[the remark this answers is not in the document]
geoHeil: it was there in the meantime - but llms removed it again - sorry for that

docs/reference/api/metadata-stores/delta.md
geoHeil: done; fixed

docs/reference/api/metadata-stores/delta.md
geoHeil: removed

mkdocs.yml
```diff
@@ -192,13 +192,15 @@ nav:
...
-        - DuckDB: reference/api/metadata-stores/ibis/duckdb.md
-        - ClickHouse: reference/api/metadata-stores/ibis/clickhouse.md
-      - Memory: reference/api/metadata-stores/memory.md
+      - Metadata Stores:
+        - reference/api/metadata-stores/index.md
+        - Databases:
+          - reference/api/metadata-stores/ibis/index.md
+          - DuckDB: reference/api/metadata-stores/ibis/duckdb.md
+          - ClickHouse: reference/api/metadata-stores/ibis/clickhouse.md
+        - Stores without engine:
```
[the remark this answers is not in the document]
geoHeil: ok/done

src/metaxy/metadata_store/base.py
```diff
@@ -480,30 +483,79 @@ def write_metadata(
...
-                df = df.to_polars()
-            elif hasattr(df, "to_pandas"):
-                df = pl.from_pandas(df.to_pandas())
-            else:
-                raise TypeError(f"Cannot convert {type(df)} to Polars DataFrame")
+        # Convert Narwhals to native Polars (both lazy and eager)
+        if isinstance(df, (nw.LazyFrame, nw.DataFrame)):
+            df = df.to_native()  # type: ignore[assignment]
+
+        if pa is not None and isinstance(df, pa.Table):
```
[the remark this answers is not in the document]
geoHeil: so should we leave it as is for this part for now?

src/metaxy/metadata_store/base.py
geoHeil: done

src/metaxy/metadata_store/delta.py
geoHeil: removed_

src/metaxy/metadata_store/delta.py
```diff
@@ -0,0 +1,484 @@
...
+        if self._local_root_path is not None:
+            return self._list_features_from_local_root()
+        return self._list_features_from_object_store()
+
+    def _list_features_from_local_root(self) -> list[FeatureKey]:
+        if self._local_root_path is None or not self._local_root_path.exists():
+            return []
+
+        feature_keys: list[FeatureKey] = []
+        for delta_log_dir in self._local_root_path.rglob("_delta_log"):
```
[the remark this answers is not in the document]
geoHeil: removed

src/metaxy/metadata_store/delta.py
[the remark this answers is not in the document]
danielgafni: I think we should either do it - in a separate PR - or raise a NotImplementedError for DeltaLake. Perhaps you can just raise right now. 

We should really re-consider this functionality. It might not be obvious how to implement it even for store types that support some kind of list operation. We will have to tag our tables (to separate from non-metaxy tables) somehow. 

Right now I don't think it should be part of the core Metaxy functionality.

src/metaxy/metadata_store/base.py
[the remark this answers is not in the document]
danielgafni: Yes, sure

src/metaxy/metadata_store/base.py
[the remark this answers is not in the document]
geoHeil: it removed it again - reworking

src/metaxy/metadata_store/delta.py
```diff
@@ -0,0 +1,406 @@
...
+            )
+            self._delta_table_cache[cache_key] = table
+        return table
+
+    # ===== Storage operations =====
+
+    def _write_metadata_impl(
+        self,
+        feature_key: FeatureKey,
+        df: pl.DataFrame | pl.LazyFrame | nw.DataFrame[Any] | nw.LazyFrame[Any],
```
geoHeil: https://github.com/anam-org/metaxy/issues/280

src/metaxy/metadata_store/base.py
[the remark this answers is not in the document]
geoHeil: https://github.com/anam-org/metaxy/issues/280 for later to refine

src/metaxy/metadata_store/base.py
```diff
@@ -494,30 +494,57 @@ def write_metadata(
...
-            # Must be some other type - shouldn't happen but handle defensively
-            if hasattr(df, "to_polars"):
-                df = df.to_polars()
-            elif hasattr(df, "to_pandas"):
-                df = pl.from_pandas(df.to_pandas())
-            else:
-                raise TypeError(f"Cannot convert {type(df)} to Polars DataFrame")
+        # Convert Narwhals to native Polars (both lazy and eager)
+        if isinstance(df, (nw.LazyFrame, nw.DataFrame)):
+            df = df.to_native()  # type: ignore[assignment]
```
danielgafni: This won't convert to polars if the frame is backed by let's say Ibis. 

You have to `.lazy().collect().to_polars()`. We already have a helper for that in _utils.py if I recall correctly. 

It's also best if we issue a user warning in case we need to collect a DataFrame like this (should only happen for lazy frames not backed by Polars). 

---

I almost got this logic into main anyway, so perhaps don't bother right now, it will be fixed anyway.

src/metaxy/metadata_store/base.py
```diff
@@ -1893,13 +2006,9 @@ def _resolve_update_native(
                 )
             else:
                 return Increment(
```
danielgafni: Why are we collecting to polars here? I'd prefer to preserve the actual backend of the metadata store. 

Users can always convert to polars if needed.

src/metaxy/metadata_store/base.py
```diff
@@ -1988,18 +2118,14 @@ def _resolve_update_polars(
...
                 changed=changed
                 if isinstance(changed, nw.LazyFrame)
-                else nw.from_native(changed),
+                else nw.from_native(added),
                 removed=removed
                 if isinstance(removed, nw.LazyFrame)
                 else nw.from_native(removed),
             )
         else:
             return Increment(
```
danielgafni: Another one?

src/metaxy/metadata_store/delta.py
```diff
@@ -0,0 +1,407 @@
...
+                    f"Flat layout expects single-segment directories, got {relative_path}"
+                )
+            return FeatureKey(parts[0].split("__"))
+        elif self.layout == "nested":
+            return FeatureKey(list(relative_path.parts))
+        else:
+            raise ValueError(f"Unknown layout: {self.layout}")
+
+    @property
+    def _deltalake(self):  # type: ignore[no-untyped-def]
```
danielgafni: What's that for?

src/metaxy/metadata_store/delta.py
```diff
@@ -0,0 +1,407 @@
...
+            feature_version=feature_version,
+            filters=filters,
+            columns=columns,
+        )
+
+    def display(self) -> str:
+        """Return human-readable representation of the store."""
+        details = [f"path={self.root_path}", f"layout={self.layout}"]
+        if self.storage_options:
+            details.append("storage_options=***")
```
danielgafni: I don't see any value in this string

docs/reference/api/metadata-stores/delta.md
```diff
@@ -0,0 +1,35 @@
...
+It keeps one Delta table per feature and appends new metadata versions.
+Polars is used to write the tables.
+
+## Installation
+
+The backend relies on [`deltalake`](https://delta-io.github.io/delta-rs/python/)
+which ships with Metaxy’s `delta` dependency group:
+
+```bash
+uv sync --extra delta
```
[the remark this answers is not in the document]
geoHeil: cleaned/ we previously had it at otehr stores - I think fixed there also now

docs/reference/api/metadata-stores/delta.md
```diff
@@ -0,0 +1,35 @@
+---
+title: Delta Lake Metadata Store
+---
+
+# DeltaMetadataStore
+
+Deltalake is a object-store-backed metadata backend.
```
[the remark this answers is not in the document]
geoHeil: changed

src/metaxy/models/feature.py
```diff
@@ -894,6 +894,15 @@ class BaseFeature(FrozenBaseModel, metaclass=MetaxyMeta, spec=None):
     graph: ClassVar[FeatureGraph]
     project: ClassVar[str]
 
+    def __init_subclass__(
```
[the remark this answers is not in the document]
geoHeil: removed

src/metaxy/metadata_store/delta.py
geoHeil: cleaned

docs/reference/api/metadata-stores/delta.md
geoHeil: move to learn

src/metaxy/metadata_store/delta.py
```diff
@@ -0,0 +1,486 @@
...
+
+        Raises:
+            TableNotFoundError: If table doesn't exist and auto_create_tables is False
+        """
+        import deltalake  # pyright: ignore[reportMissingImports]
+
+        from metaxy._utils import collect_to_polars
+
+        # Convert Narwhals DataFrame to Polars for delta-rs
+        df_polars = collect_to_polars(df)
```
danielgafni: This will collect lazy frames, we need a slightly different thing: convert to Polars without collecting lazy frames. There is another helper in _utils.py that does that.

danielgafni (changes requested):
I looked through 50% but it seems you did not push your changes? my comments are marked as resolved in the UI but they are really not

src/metaxy/metadata_store/delta.py
```diff
@@ -0,0 +1,486 @@
...
+    _auto_collect_lazy_frames = False  # Handle lazy frames for streaming writes
+
+    def __init__(
+        self,
+        root_path: str | Path,
+        *,
+        storage_options: dict[str, Any] | None = None,
+        fallback_stores: list[MetadataStore] | None = None,
+        layout: str = "flat",
+        streaming_chunk_size: int | None = None,
```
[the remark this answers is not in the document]
geoHeil: fixed

docs/learn/metadata-stores/deltalake.md
```diff
@@ -0,0 +1,32 @@
+# DeltaMetadataStore
+
+Delta Lake is a object-store-backed metadata backend.
+It keeps one Delta table per feature, appends new metadata versions, and leans on Polars / Narwhals for provenance calculations.
```
danielgafni: > appends new metadata versions

this doesn't sound good

docs/learn/metadata-stores/deltalake.md
```diff
@@ -0,0 +1,32 @@
...
+
+The backend relies on [`deltalake`](https://delta-io.github.io/delta-rs/python/)
+which ships with Metaxy’s `delta` dependency group.
+
+## Object Stores
+
+Point `root_path` at any supported URI (`s3://`, `abfss://`, `gs://`, `file://`,
+`local://`, …) and forward credentials with `storage_options`.
+The dict is passed
+verbatim to `deltalake`, so you can reuse the same keys you would use with delta-rs directly.
```
danielgafni: either name it `deltalake` or `delta-rs` (I'd prefer `deltalake` since it's the Python package that's called under the hood - pretty sure Polars just does that instead of using the Rust SDK), two different names for the same thing may be confusing for users who don't know much about the ecosystem. 

Also, either a reference to the corresponding docs or a link to our own API docs would be useful: users could immediately learn more about these options by going there

src/metaxy/metadata_store/delta.py
```diff
@@ -0,0 +1,336 @@
...
+
+        Args:
+            mode: Access mode for this connection session (accepted for consistency but not used).
+
+        Yields:
+            Self: The store instance with connection open
+        """
+        # Mark store as open and validate
+        # Note: Delta auto-creates tables on first write, no need to pre-create them
+        self._is_open = True
```
danielgafni: Is this `._is_open` even used anywhere?

src/metaxy/metadata_store/delta.py
```diff
@@ -0,0 +1,336 @@
...
+
+        try:
+            yield self
+        finally:
+            # delta-rs uses one-shot operations, so nothing to close
+            self._is_open = False
+
+    # ===== Internal helpers =====
+
+    def _table_name_to_feature_key(self, table_name: str) -> FeatureKey:
```
danielgafni: Old method, I don't think we need it anymore?

src/metaxy/metadata_store/delta.py
```diff
@@ -0,0 +1,336 @@
...
+        return os.path.join(self._root_uri, table_path)
+
+    def _table_exists(self, table_uri: str) -> bool:
+        """Check whether the provided URI already contains a Delta table.
+
+        Works for both local and remote (object store) paths.
+        """
+        import deltalake  # pyright: ignore[reportMissingImports]
+        from deltalake.exceptions import (
+            TableNotFoundError,  # pyright: ignore[reportMissingImports]
```
danielgafni: This may clash with our own `TableNotFoundError`, maybe import it `as DeltaTableNotFoundError`? not that it's important, just would be easier to read the code

src/metaxy/metadata_store/delta.py
```diff
@@ -0,0 +1,336 @@
...
+        if columns is not None:
+            nw_lazy = nw_lazy.select(columns)
+
+        return nw_lazy
+
+    def display(self) -> str:
+        """Return human-readable representation of the store."""
+        details = [f"path={self._display_root}"]
+        if self.storage_options:
+            details.append("storage_options=***")
```
danielgafni: again what's the value of this string? I don't think it does anything useful.
At least display the keys?

geoHeil:
there should now be tests with moto

src/metaxy/metadata_store/delta.py
```diff
@@ -0,0 +1,336 @@
...
+        df: Frame,
+        **kwargs: Any,
+    ) -> None:
+        """Append metadata to the Delta table for a feature.
+
+        Args:
+            feature_key: Feature key to write to
+            df: DataFrame with metadata (already validated)
+            **kwargs: Write options:
+                - streaming_chunk_size (int | None): Override constructor's streaming_chunk_size
```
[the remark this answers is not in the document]
danielgafni: Didn't realize that

tests/metadata_stores/conftest.py
```diff
@@ -255,6 +255,13 @@ def case_clickhouse(
     ) -> tuple[type[MetadataStore], dict[str, Any]]:
         return (ClickHouseMetadataStore, {"connection_string": clickhouse_db})
 
+    def case_delta(
```
geoHeil: move to AllStoreCases

tests/metadata_stores/conftest.py
```diff
@@ -255,6 +254,13 @@ def case_clickhouse(
...
```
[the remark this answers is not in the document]
geoHeil: what about the other onesin StoreCases? can we remove them?

danielgafni (changes requested):
Very nice! Now it's a clean PR. Please un-hardcode the port and we can merge

tests/metadata_stores/conftest.py
```diff
@@ -428,3 +447,46 @@ def test_something(config_with_truncation):
     # Set and restore config
     with config.use():
         yield config
+
+
+@pytest.fixture(scope="session")
+def s3_endpoint_url() -> Generator[str, None, None]:
+    """Start a moto S3 server on a random free port."""
+    port = find_free_port()
```
danielgafni: Specifically with `ThreadedMotoServer` this is not needed because it has native functionality for this: 

```py
server = ThreadedMotoServer(port=0)
server.start()
host, port = server.get_host_and_port()
```

But it doesn't matter.

PR #237, opened 2025-11-05
lancedb store

### LINKED ISSUE #233: [store] lancedb

geoHeil:
resolves #233 

## details

https://lancedb.com/docs/tables/schema/

lance offers 2D data + schema migration - we may want to explore if we can make use of that

see also https://www.youtube.com/watch?v=4iKAOCw-_AA

https://www.youtube.com/watch?v=VAO4407nHTM https://lancedb.com/docs/geneva/

Their backfill https://lancedb.com/docs/geneva/jobs/backfilling/ has the benefit of not replicating the BLOBs


src/metaxy/metadata_store/lancedb.py
```diff
@@ -0,0 +1,333 @@
...
+            ConnectionError: If remote connection fails (e.g., invalid credentials)
+        """
+        import lancedb
+
+        # Only create directory for local filesystem paths
+        # Remote URIs (s3://, db://, http://, https://, gs://, etc.) are handled by LanceDB
+        if not self._is_remote_uri(self.database_path):
+            Path(self.database_path).mkdir(parents=True, exist_ok=True)
+
+        self._conn = lancedb.connect(self.database_path)
```
geoHeil: Would we want to enable async mode? 

https://lancedb.com/docs/integrations/platforms/polars_arrow/

src/metaxy/metadata_store/lancedb.py
```diff
@@ -0,0 +1,333 @@
...
+            Narwhals LazyFrame with metadata, or None if table not found
+        """
+        self._check_open()
+        feature_key = self._resolve_feature_key(feature)
+        table_name = self._table_name(feature_key)
+        if not self._table_exists(table_name):
+            return None
+
+        table = self._get_table(table_name)
+        pl_lazy = table.to_polars()
```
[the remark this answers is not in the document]
geoHeil: we currently try/catch this and fall back to eager mode when loading.

Q: should we really downgrade our polars minimal requirement here? I think this is perhaps even cleaner (though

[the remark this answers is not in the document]
danielgafni: yeah do not handle this exception, it's not our problem. downgrade our min polars version if there is a bug in lancedb with the recent versions.

src/metaxy/metadata_store/lancedb.py
```diff
@@ -0,0 +1,350 @@
...
+            if "batch_size" not in str(exc):
+                raise
+            logger.debug(
+                "Polars/LanceDB batch_size incompatibility hit; converting via Arrow "
+            )
+            # Fall back to eager Arrow conversion until LanceDB issue #1539 is resolved.
+            arrow_table = table.to_arrow()
+            pl_lazy = pl.DataFrame(arrow_table).lazy()
+
+        nw_lazy = nw.from_native(pl_lazy)
```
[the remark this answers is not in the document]
geoHeil: polars 1.3.0 should support it - currently 1.35.2 is latest reason is https://github.com/lancedb/lancedb/issues/1539

No we cannot:

and polars-hash>=0.5.1,<=0.5.4 depends on polars>=1.5.0, 

there would be a version conflict

we could not do the try/catch and then just support eager?

geoHeil:
> Thanks, looking promising!
> 
> Not sure if we need to add the `auto_create_tables` parameter.. does lancedb handle schema evolution out of the box like deltalake? in this case it's safe to always auto-create them

did you see 

https://lancedb.com/docs/tables/schema/

lance offers 2D data + schema migration - we may want to explore if we can make use of that

see also https://www.youtube.com/watch?v=4iKAOCw-_AA

https://www.youtube.com/watch?v=VAO4407nHTM https://lancedb.com/docs/geneva/

src/metaxy/metadata_store/lancedb.py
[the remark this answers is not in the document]
danielgafni: Can't we also downgrade polars-hash? 

I recall the was a compatibility issue with an older version of it, so maybe not. 

Well.. at least let's handle it with a warning if there is no other option to get it working at all.

src/metaxy/metadata_store/lancedb.py
```diff
@@ -0,0 +1,350 @@
...
+
+            # LanceDB Cloud (requires LANCEDB_API_KEY environment variable)
+            cloud = LanceDBMetadataStore("db://my-database")
+
+            # Local with fallback to production
+            dev = LanceDBMetadataStore("./dev", fallback_stores=[prod])
+            ```
+        """
+        # Store as string to handle both local paths and remote URIs
+        self.database_path = str(database_path)
```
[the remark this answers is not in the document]
geoHeil: Don`t they use URI?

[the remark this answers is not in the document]
geoHeil: usng URI now

tmp/migrations/20251118_193558.yaml
```diff
@@ -0,0 +1,8 @@
+migration_type: metaxy.migrations.models.DiffMigration
```
geoHeil: migration fails unless this additional metadata is specified

docs/learn/metadata-stores/lancedb.md
```diff
@@ -0,0 +1,175 @@
+# LanceDBMetadataStore
+
+## Installation
+
+Install the [`lancedb`](https://lancedb.com/) backend alongside Metaxy using the `lancedb` dependency group.
```
geoHeil: extras, not dependency group

docs/learn/metadata-stores/lancedb.md
```diff
@@ -0,0 +1,17 @@
...
+
+## Installation
+
+The backend relies on [`lancedb`](https://lancedb.com/) which ships with Metaxy's `lancedb` extras.
+
+## Object Stores
+
+Point `uri` at any supported URI (`s3://`, `gs://`, `az://`, `db://`, …) and forward credentials with the platform's native mechanism (environment variables, IAM roles, workload identity, etc.). LanceDB supports local filesystem, S3, GCS, Azure, LanceDB Cloud, and remote HTTP/HTTPS endpoints.
+
+## Storage Layout
```
danielgafni: no nested?

docs/learn/metadata-stores/lancedb.md
```diff
@@ -0,0 +1,17 @@
...
+
+- Each feature is stored as an individual Lance table: `{namespace}__{feature_name}`
+- System tables: `metaxy-system__feature_versions`, `metaxy-system__migration_events`
```
danielgafni: don't need to mention them here

src/metaxy/metadata_store/lancedb.py
```diff
@@ -0,0 +1,604 @@
...
+    """
+
+    _should_warn_auto_create_tables = False
+
+    def __init__(
+        self,
+        uri: str | Path,
+        *,
+        fallback_stores: list[MetadataStore] | None = None,
+        auto_create_tables: bool | None = True,
```
danielgafni: isn't it always True for us? the argument is not needed?

src/metaxy/metadata_store/lancedb.py
```diff
@@ -0,0 +1,604 @@
...
+
+        # No scheme means it's a filesystem path (relative or absolute)
+        return True
+
+    def _table_name(self, feature_key: FeatureKey) -> str:
+        return feature_key.table_name
+
+    def _table_exists(self, table_name: str) -> bool:
+        assert self._conn is not None, "Store must be open"
+        return table_name in self._conn.table_names()  # type: ignore[attr-defined]
```
danielgafni: use `self.conn` property

tests/metadata_stores/conftest.py
```diff
@@ -375,6 +376,18 @@ def case_delta(self, tmp_path: Path) -> MetadataStore:
             hash_algorithm=HashAlgorithm.XXHASH64,
         )
 
+    @pytest.mark.lancedb
+    @pytest.mark.polars
+    def case_lancedb(self, tmp_path: Path) -> MetadataStore:
+        pytest.importorskip("lancedb")
```
danielgafni: not a fan

src/metaxy/metadata_store/lancedb.py
```diff
@@ -0,0 +1,604 @@
...
+        # LanceDB supports both Polars DataFrames and Arrow tables directly
+        # Try Polars first (native integration), fall back to Arrow if needed
+        try:
+            if self._table_exists(table_name):
+                table = self._get_table(table_name)
+                # Use Polars DataFrame directly - LanceDB handles conversion
+                table.add(df_polars)  # type: ignore[attr-defined]
+            else:
+                # Create table from Polars DataFrame - LanceDB handles schema
+                self._conn.create_table(table_name, data=df_polars)  # type: ignore[attr-defined]
```
danielgafni: can we use the pydantic integration for creating tables? 
this might be a little challenge because currently system tables are not features (and that's correct) and we would need another way to convert a system table key into the class

docs/learn/metadata-stores/lancedb.md
geoHeil: Compared to i.e. Postgres  (db, scema, table) you would have  db and table where db more or less maps to schema.

So as this all sits on S3 we shoudl clarify if we have one `s3:///some/folder/<<various_feature_dbs>>/some_feature_tables` or if we have `s3:///some/folder/metaxy.lance/some_feature_tables` where `some_feature_tables` refers to Lance internally managed tables

geoHeil: TLDR: We could nest on the prefix of the DB but I am not sure if we want that right now

src/metaxy/metadata_store/lancedb.py
geoHeil: But if we do that it should be generic/plain pydantic and not LanceModel right?

Would it make sense to do it like protobuf/grpc and just have some conde generating process and then these concrete instances are used at runtime?

For the non-system tables: The pydantic models also lack the system columns and we would need to monkey-patch them.


How about a:

```python
def create_lance_model_for_feature(feature_cls: type[BaseFeature]) -> type[LanceModel]:
      """Auto-generate LanceModel from BaseFeature + system columns.
      
      Combines user-defined fields (from BaseFeature) with Metaxy system columns.
      """
      # Get user's fields from Pydantic BaseFeature
      user_fields = {
          name: (field.annotation, field.default if field.is_required() else None)
          for name, field in feature_cls.model_fields.items()
      }

      # Add Metaxy system columns (consistent across all features)
      system_fields = {
          'metaxy_provenance_by_field': (dict, ...),
          'metaxy_feature_version': (str, ...),
          'metaxy_snapshot_version': (Optional[str], None),
          'metaxy_feature_hash': (Optional[str], None),
      }

      # Combine and create LanceModel
      all_fields = {**user_fields, **system_fields}
      return create_model(
          f"{feature_cls.__name__}Metadata",
          __base__=LanceModel,
          **all_fields
      )
  lance_model = create_lance_model_for_feature(VideoFeature)
```

docs/reference/api/metadata-stores/index.md
```diff
@@ -12,6 +12,7 @@ Metaxy has built-in support for the following metadata store types:
 ## Storage Only
 
 - [DeltaMetadataStore](./delta.md).
+- [LanceDBMetadataStore](./lancedb.md).
```
danielgafni: LanceDB isn't a storage-only DB, right? similar to ducklake in this sense. 

We should put it into "databases" and explain that it can be used in both embedded and external compute variants.

src/metaxy/metadata_store/lancedb.py
danielgafni: I think `create_lance_model_for_feature`​ is reasonable approach, and we could also make an integration similar to `sqlmodel`​, but let's park both options for later

src/metaxy/metadata_store/lancedb.py
```diff
@@ -0,0 +1,438 @@
...
+        # Increment context depth to support nested contexts
+        self._context_depth += 1
+
+        try:
+            # Only perform actual open on first entry
+            if self._context_depth == 1:
+                import lancedb
+
+                # Only create directory for local filesystem paths
+                # Remote URIs (s3://, db://, http://, https://, gs://, etc.) are handled by LanceDB
```
danielgafni: that's a weird comment, object stores do not have directories

src/metaxy/metadata_store/lancedb.py
```diff
@@ -0,0 +1,438 @@
...
+        if "://" in path:
+            return False
+
+        # No scheme means it's a filesystem path (relative or absolute)
+        return True
+
+    def _table_name(self, feature_key: FeatureKey) -> str:
+        return feature_key.table_name
+
+    def _table_exists(self, table_name: str) -> bool:
```
danielgafni: lanceb doesn't have a native `.table_exists()`​ method?

src/metaxy/metadata_store/lancedb.py
```diff
@@ -0,0 +1,438 @@
...
+            return None
+
+        table = self._get_table(table_name)
+
+        try:
+            # shim for https://github.com/lancedb/lancedb/issues/1539
+            pl_lazy = table.to_polars()
+            # Smoke-test the LazyFrame to surface the batch_size issue eagerly.
+            pl_lazy.limit(0).collect()
+        except TypeError as exc:
```
danielgafni: can we remove this hand-rolled sink implementation?

tests/metadata_stores/test_lancedb.py
```diff
@@ -0,0 +1,286 @@
...
+    # Should raise when accessing conn while closed
+    with pytest.raises(StoreNotOpenError, match="LanceDB connection is not open"):
+        _ = store.conn
+
+    with store:
+        conn = store.conn
+        assert conn is not None
+
+
+def test_lancedb_is_local_path() -> None:
```
danielgafni: should just have this shared with deltalake

danielgafni (changes requested):
Great! 

1. Minor comments
2. generic functionality like url masking and checking whether if a path is local should be shared across stores. The second functionality already exists in deltalake. You can refactor this in a new downstack (previous) PR. 
3. Needs a test with moto

tests/metadata_stores/test_lancedb.py
geoHeil: we should also have it shared with PG (for the display/sanitize)

danielgafni (changes requested):
Let's think about this question I've left before we merge this, I believe evething else should be ready!

src/metaxy/metadata_store/lancedb.py
```diff
@@ -0,0 +1,363 @@
...
+        self._check_open()
+        feature_key = self._resolve_feature_key(feature)
+        table_name = self._table_name(feature_key)
+        if not self._table_exists(table_name):
+            return None
+
+        table = self._get_table(table_name)
+        # https://github.com/lancedb/lancedb/issues/1539
+        # Fall back to eager Arrow conversion until LanceDB issue #1539 is resolved.
+        arrow_table = table.to_arrow()
```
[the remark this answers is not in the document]
danielgafni: Datafusion? :scream:  
  
https://lance.org/integrations/datafusion/#python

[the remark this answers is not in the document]
geoHeil: > It looks like we could use duckdb instead? https://lance.org/integrations/duckdb/

they write:

- A LanceDataset is accessible to DuckDB through the Arrow compatibility layer directly (given it uses arrow how would this be different from the polars --> arrow --> lance)?
-  can also be queried  (hopefully the write path also works)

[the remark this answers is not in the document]
danielgafni: Yeah having it lazy would probably be very important for production (lower memory requirements). 

You are right about DuckDB, I didn't notice it would use arrow as well. 

Alright, looks like we can't avoid it until the polars bug is fixed.

PR #254, opened 2025-11-07
allow python migration files

### LINKED ISSUE #49: [core] Allow Python migration files
Add support for Python migration files alongside or instead of the current YAML-based migrations.

Python migrations would enable more complex migration logic that YAML cannot express, such as conditional logic, data transformations, or custom validation.

## Acceptance Criteria
- [ ] Support `.py` migration files in addition to `.yaml`
- [ ] Define Python migration API/interface (base class or protocol)
- [ ] Allow mixing Python and YAML migrations in the same migration directory
- [ ] Ensure Python migrations are idempotent and recoverable
- [ ] Support same operations as YAML (DataVersionReconciliation, etc.)
- [ ] Add ability to convert simple Python migrations to YAML for documentation
- [ ] Update migration discovery to find both file types
- [ ] Add tests for Python migration execution
- [ ] Document when to use Python vs YAML migrations

## Example Structure
```python
from metaxy.migrations import Migration, DataVersionReconciliation

class Migration_2024_01_15_refactor_features(Migration):
    from_snapshot_version = "abc123"
    to_snapshot_version = "def456"
    
    def operations(self):
        return [
            DataVersionReconciliation(
                feature_key="my_feature",
                reason="Refactored computation logic"
            ),
            # Custom logic here
        ]
```

geoHeil:
resolves #49


danielgafni:
I won't be looking into this yet, this feature is not very important

danielgafni:
@geoHeil are you still working on this or can I close it? I don't think we need this any time soon

geoHeil:
lets close it

PR #287, opened 2025-11-16
Revert "Remove redundant tests"

danielgafni:
Reverts anam-org/metaxy#231


danielgafni:
cc @geoHeil - please open a new one that would only *remove* tests, and also avoid specifying features outside of a temporary graph context (it can lead to dirty feature graphs)

PR #291, opened 2025-11-17
[core] deduplicate records in MetadataStore.read_metadata by metaxy_created_at

### LINKED ISSUE #243: [core] Support (finally) resolving increments without code_version change
This is mostly useful for root features that take arbitrary data as input. 
Should be as simple as adding as looking at `metaxy_created_at` for records when loading them and picking the latest one. 

Then the following workflow should work: 
- the user updates the root feature (by appending new metadata)
- `.resolve_update()` for a child feature now should load all upstream records (they will have a matching `code_version` -> `feature_version` and filter out the latest ones.

danielgafni:
Resolve #243

Resolve DAT-13

### DIFF
--- a/src/metaxy/metadata_store/base.py
+++ b/src/metaxy/metadata_store/base.py
@@ -10,7 +10,7 @@ from collections.abc import Iterator, Mapping, Sequence
 from contextlib import AbstractContextManager, contextmanager
 from datetime import datetime, timezone
 from types import TracebackType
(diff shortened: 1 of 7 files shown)


danielgafni:
### Merge activity

* **Nov 18, 2:45 PM UTC**: @danielgafni merged this pull request with [Graphite](https://app.graphite.com/github/pr/anam-org/metaxy/291).

PR #331, opened 2025-11-21
[sqlalchemy] create Sqlalchemy plugin

danielgafni:


src/metaxy/ext/alembic/plugin.py
```diff
@@ -0,0 +1,200 @@
...
+
+        # Load features
+        init_metaxy()
+
+        # Get metadata for current project only
+        target_metadata = get_features_sqlalchemy_metadata()
+        ```
+    """
+    from sqlalchemy import MetaData
+    from sqlmodel import SQLModel
```
danielgafni: Same, we should not depend on SQLModel.

Issue #334, opened 2025-11-21
[core] “filters” field to FeatureDep


geoHeil:
implement: “filters” field to FeatureDep that would use the  SQL->narwhals parser

This would be for static filters

sql_filters: Sequence[str]
But use a serialization name “filters”
Then make a @cached_propety “filters” that return the parsed narwhals expressions from sql_filters

PR #335, opened 2025-11-22
filters field to FeatureDep

geoHeil:
resolves https://github.com/anam-org/metaxy/issues/334

### DIFF
--- a/src/metaxy/models/feature_spec.py
+++ b/src/metaxy/models/feature_spec.py
@@ -4,14 +4,9 @@ import hashlib
 import json
 from collections.abc import Mapping, Sequence
 from functools import cached_property
-from typing import (
-    TYPE_CHECKING,
-    Annotated,
-    Any,
-    TypeAlias,
-    overload,
-)
+from typing import TYPE_CHECKING, Annotated, Any, TypeAlias, overload
 
+import narwhals as nw
 import pydantic
 from pydantic import BeforeValidator
 from pydantic.types import JsonValue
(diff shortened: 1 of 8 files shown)


geoHeil:
* **#335** <a href="https://app.graphite.com/github/pr/anam-org/metaxy/335?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anam-org/metaxy/335?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

danielgafni (approved):
Very nice! 

Let's merge this since this is already a working solution. 

---

In the future, I would like to explore the following changes:

- change `filters` from `@cached_property` to a normal Pydantic attribute 
- allow arbitrary types so we can set this field directly with Narwhals expressions from Python 
- add validators that would ensure that *either* `sql_filters` *or* `filters` is provided to the model at instantiation time
- somehow convert `filters` to `sql_filters` for serialization, I am not sure if we can do it right now, we only have str->narwhals parsing and not the other way, right? Maybe we could also start without this and just log a warning mentioning how `.filters` set directly would be lost after serialization. Not sure if that's a good idea. Probably not. 

This would make it easier to set `FeatureDep.filters` from Python.

danielgafni (changes requested):
Uhh I was too fast. We need at least one integration test with a metadata store using this feature. The current tests don't guarantee that metadata stores eventually correctly invoke `FeateureDepTransformer` to do it.

danielgafni (changes requested):
I don't like the addition of direct filters field on this PR, my comment explicitly mentioned I wanted to explore it later. 

I also mentioned the warning thing might be dangerous. 

This will lead to data loss if filters are set from Python, I don't think we should allow it. We should first explore it it's possible to serialize narwhals expressions into our sql strings. If that's possible we could always make sure the serialized filters are correct. 

Could you revert the implementation to what I've approved originally? 

P.S. I think the validation code can be simplified greatly. We only need a ValidatedNarwhalsFilter Annotated type which would already specify the fight Pydantic validator. We then would get tuple[ValidatedNarwhalsFilter, ...] (or lisf, or any other nested types) parsing out of the box.

PR #381, opened 2025-11-26
[core] FlatVersioningEngine

geoHeil:


danielgafni (changes requested):
In principle this is doing the right thing, but we need better architecture to avoid that much code duplication. 

This basically re-implements the entire versioning engine. 

Maybe let's think if it's possible to only alter some code paths instead of redoing everything from scratch? We need to share code to reduce risks in such a core component.

src/metaxy/versioning/dict_based.py
```diff
@@ -0,0 +1,94 @@
...
+        field_name: str,
+    ) -> nw.Expr:
+        """Access a field from a virtual struct column."""
+        flattened_name = FlatVersioningEngine._get_flattened_column_name(
+            struct_column, field_name
+        )
+        return nw.col(flattened_name)
+
+
+class DictBasedVersioningEngine(FlatVersioningEngine, VersioningEngine, ABC):
```
danielgafni: It seems like this is supposed to be named FlatVersioningEngine?

src/metaxy/versioning/dict_based.py
```diff
@@ -0,0 +1,94 @@
...
+class FlatVersioningEngine:
+    """Helpers for representing struct-like data as flattened columns."""
+
+    @staticmethod
+    def _get_flattened_column_name(struct_name: str, field_name: str) -> str:
+        """Get the flattened column name for a struct field."""
+        return f"{struct_name}__{field_name}"
+
+    @staticmethod
+    def build_struct_column(
```
danielgafni: It's weird how a method that's called build_struct_column doesn't build any struct columns 

Should it be renamed? And I mean renamed everywhere, on the base class as well.

src/metaxy/versioning/dict_based.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+class DictBasedVersioningEngine(FlatVersioningEngine, VersioningEngine, ABC):
+    """Base class for versioning engines that use dict-based struct representation."""
+
+    def __init__(self, plan: FeaturePlan):
+        VersioningEngine.__init__(self, plan)
+
+
+class IbisDictBasedVersioningEngine(FlatVersioningEngine, IbisVersioningEngine):
```
danielgafni: IbisFlatVersioningEngine

src/metaxy/versioning/engine.py
```diff
@@ -461,9 +493,20 @@ def hash_struct_version_column(
         field_names = sorted([f.key.to_struct_key() for f in self.plan.feature.fields])
 
         # Concatenate all field hashes with separator
-        sample_components = [
-            nw.col(struct_column).struct.field(field_name) for field_name in field_names
-        ]
+        sample_components = []
+        for field_name in field_names:
+            expr = self.access_provenance_field(struct_column, field_name)
+            # If data_version is missing, fall back to provenance-by-field (useful for dict-based stores)
```
danielgafni: I'm not sure we are supposed to be handling this, if it's missing something went really wrong. 

write_metadata always ensures this column is present so it can't happen. 

In tests it's our responsibility to provide it. We have a hypothesis strategy that does it.

danielgafni (changes requested):
Alright, I think the implementation is good, but this cannot be merged until the docstrings are cleaned up (I'd encourage you to remove unnecessary docstrings, especially from trivial methods). 

And please rename dict to flat **everywhere**, not just the (ex) mixin.

src/metaxy/versioning/dict_based.py
geoHeil: it is bulding a struct in all other cases - here we are just doing something else/compatible for stores which are not able to build the struct.

It would be cleaner if we find a better super-name

geoHeil: how about `record_field_versions`?

src/metaxy/metadata_store/base.py
```diff
@@ -1280,12 +1344,29 @@ def _add_system_columns(
...
         if METAXY_PROVENANCE_BY_FIELD not in df.columns:
-            raise ValueError(
-                f"Metadata is missing a required column `{METAXY_PROVENANCE_BY_FIELD}`. It should have been created by a prior `MetadataStore.resolve_update` call. Did you drop it on the way?"
-            )
+            # Build struct from flattened provenance columns if available (dict-based stores)
+            prov_flat = [
+                c for c in df.columns if c.startswith(f"{METAXY_PROVENANCE_BY_FIELD}__")
+            ]
+            if prov_flat:
+                struct_fn: Any = getattr(nw, "struct")
```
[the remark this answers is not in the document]
geoHeil: cleaned up

src/metaxy/metadata_store/base.py
```diff
@@ -602,6 +614,58 @@ def resolve_update(
...
 
+        # Allow subclasses to post-process the result (e.g., for type conversions)
+        return self._post_process_resolve_update_result(result, lazy=lazy)
+
+    def _preprocess_samples_for_resolve_update(
+        self,
+        *,
+        feature: type[BaseFeature],
+        plan: FeaturePlan,
+        samples: Frame | None,
```
danielgafni: Maybe just name it `df` and use `FrameT`

src/metaxy/metadata_store/base.py
```diff
@@ -602,6 +614,58 @@ def resolve_update(
...
+        filters: Mapping[FeatureKey, Sequence[nw.Expr]] | None,
+        lazy: bool,
+    ) -> Frame | None:
+        """Hook for stores to adjust user-provided samples before resolve_update.
+
+        Default implementation returns samples unchanged.
+        """
+        return samples
+
+    def _post_process_resolve_update_result(
```
danielgafni: Needs overloads for lazy and eager variants

src/metaxy/versioning/engine.py
```diff
@@ -457,13 +474,13 @@ def hash_struct_version_column(
...
         field_names = sorted([f.key.to_struct_key() for f in self.plan.feature.fields])
 
         # Concatenate all field hashes with separator
-        sample_components = [
-            nw.col(struct_column).struct.field(field_name) for field_name in field_names
-        ]
+        sample_components = []
+        for field_name in field_names:
+            expr = self.access_provenance_field(struct_column, field_name)
+            sample_components.append(expr.fill_null(""))
```
danielgafni: I guess the fill_null is for cases when the set of fields has changed?

src/metaxy/metadata_store/base.py
```diff
@@ -602,6 +614,58 @@ def resolve_update(
...
             )
 
+        # Allow subclasses to post-process the result (e.g., for type conversions)
+        return self._post_process_resolve_update_result(result, lazy=lazy)
+
+    def _preprocess_samples_for_resolve_update(
+        self,
+        *,
+        feature: type[BaseFeature],
+        plan: FeaturePlan,
```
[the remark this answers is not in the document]
geoHeil: Are you sure it is lossless to convert?

danielgafni (dismissed):
Another `Frame` you've missed

---

I really don't want to be talking to unfiltered AI, that's a waste of time

src/metaxy/metadata_store/base.py
```diff
@@ -229,6 +245,157 @@ def __init__(
 
         self.fallback_stores = fallback_stores or []
 
+    @contextmanager
+    def _with_feature_plan(self, plan: FeaturePlan) -> Iterator[None]:
```
danielgafni: This is super odd, why don't we just pass it around as a normal variable?

geoHeil:
> Alright, I think these are some very good steps in the right direction, but let's cleanup our inheritance patterns.
> 
> Flat and Struct Ibis engines should have the same inheritance patterns. But rignt now one of them inherits from a mixin (a mixin is typically only useful if it's used more than one time) and the other doesn't.
> 
> Let's introduce a new abstract `FieldAccessor` and move what's in the current mixin into `FlatFieldAccessor`, and the corresponding methods that now live on `IbisVersioningEngine` to `StructFieldAccessor`.
> 
> Is the plan clear?
> 
> Also, please spend more time on reviews, the LLMs are introducing a bunch of bullshit code on a reagular basis

interesting https://github.com/anam-org/metaxy/pull/381#pullrequestreview-3522810112 ...

src/metaxy/metadata_store/base.py
```diff
@@ -58,13 +59,45 @@
...
+    upstream: dict[FeatureKey, FrameT]
+    sample: FrameT | None
+
+
+@dataclass(frozen=True)
+class _NormalizedEngineInputs:
+    inputs: _EngineInputs[nw.DataFrame[Any]] | _EngineInputs[nw.LazyFrame[Any]]
+
+    @property
+    def is_polars(self) -> bool:
```
danielgafni: I'm not sure if it's a good idea to implement a method with such logic, it gives pretty vague information about the data frames... why would care if the first one is polars?

src/metaxy/metadata_store/base.py
```diff
@@ -229,6 +262,150 @@ def __init__(
 
         self.fallback_stores = fallback_stores or []
 
+    def _normalize_engine_inputs(
+        self,
+        *,
+        plan: FeaturePlan,
+        switched_to_polars: bool,
```
danielgafni: since it's not a local variable anymore, can it get a more descriptive name? it was hard to understand the meaning of this variable, what exactly has been switched to polars? we are doing more switching in the body of this function

PR #479, opened 2025-12-04
Update badge URL for prek in README.md

j178:
Thanks for adopting prek! This PR updates the prek badge URL in README.md to use the new path `docs/assets/badge-v0.json` instead of `assets/badge/v0.json`.


danielgafni:
Wow, that's what I call user support!

PR #504, opened 2025-12-08
[core] Implement `MetadataStore.delete_metadata` and hard deletions for Ibis

geoHeil:


src/metaxy/metadata_store/base.py
```diff
@@ -1453,6 +1475,72 @@ def drop_feature_metadata(self, feature: CoercibleToFeatureKey) -> None:
...
+                include_soft_deleted=False,
+                latest_only=latest_only,
+                allow_fallback=False,
+            )
+            if lazy is None:
+                raise FeatureNotFoundError(
+                    f"Feature {feature_key.to_string()} not found in store; cannot soft delete."
+                )
+
+            df = lazy.collect()
```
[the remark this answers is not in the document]
geoHeil: But we have to collect it somewhere: `df: IntoFrame,` for `write_metadata` is not lazy. So by dropping the `Deletionresult` we already got rid of a couple of collects. Is it fine to keep this one here for starters?

src/metaxy/metadata_store/base.py
```diff
@@ -1597,6 +1615,92 @@ def drop_feature_metadata(self, feature: CoercibleToFeatureKey) -> None:
...
+        current_only: bool = True,
+    ):
+        """Delete records matching provided filters.
+
+        Performs a soft delete by default. This is achieved by setting metaxy_deleted_at to the current timestamp.
+        Subsequent [[MetadataStore.read_metadata]] calls would drop these records by default.
+
+        Args:
+            feature: Feature to delete from.
+            filters: Single Narwhals expression or a sequence of expressions to combine with logical AND.
```
danielgafni: `one or more Narwhals expressions` that determine which records to delete

src/metaxy/metadata_store/base.py
```diff
@@ -1597,6 +1615,92 @@ def drop_feature_metadata(self, feature: CoercibleToFeatureKey) -> None:
...
+            soft: Whether to perform a soft delete.
+            current_only: When soft deleting, restrict to the latest feature version.
```
danielgafni: whether to only affect the records with the current feature version. Set this to `False` to also affect historical metadata.

src/metaxy/metadata_store/base.py
```diff
@@ -1597,6 +1615,92 @@ def drop_feature_metadata(self, feature: CoercibleToFeatureKey) -> None:
...
+            now = datetime.now(timezone.utc)
+
+            # Convert to native for datetime arithmetic, add 1μs, convert back
+            native_df = nw_df.to_native()
+            from typing import cast
+
+            import polars as pl
+
+            # Convert to Polars DataFrame for consistent datetime operations
+            if isinstance(native_df, pl.DataFrame):
```
danielgafni: This block should go away, we just need `lazy = lazy.with_columns(nw.col("METAXY_DELETED_AT", datetime.now(timezone.utc())`

src/metaxy/metadata_store/base.py
```diff
@@ -1597,6 +1615,92 @@ def drop_feature_metadata(self, feature: CoercibleToFeatureKey) -> None:
...
+            filter_expr = reduce(
+                lambda acc, expr: acc & expr, filter_list[1:], filter_list[0]
+            )
+
+        if soft:
+            lazy = self.read_metadata(
+                feature_key,
+                filters=[filter_expr],
+                include_soft_deleted=False,
+                latest_only=current_only,
```
danielgafni: wrong argument set, has to be `current_only=current_only`

src/metaxy/metadata_store/base.py
```diff
@@ -1597,6 +1615,92 @@ def drop_feature_metadata(self, feature: CoercibleToFeatureKey) -> None:
...
+                allow_fallback=False,
+            )
+            if lazy is None:
```
danielgafni: this check is not needed, we no longer return a `None` if the feature was not found

src/metaxy/metadata_store/base.py
```diff
@@ -1597,6 +1615,92 @@ def drop_feature_metadata(self, feature: CoercibleToFeatureKey) -> None:
...
+                raise FeatureNotFoundError(
+                    f"Feature {feature_key.to_string()} not found in store; cannot soft delete."
+                )
+
+            df = lazy.collect()
```
danielgafni: I don't see why do we need this collect.. why do we? 

We should be able to just set `METAXY_DELETED_AT` lazily and write the lazy df back. In case of Ibis this would result in `INSERT INTO SELECT FROM` query with no materialization in memory.

src/metaxy/metadata_store/delta.py
```diff
@@ -315,6 +315,16 @@ def _drop_feature_metadata_impl(self, feature_key: FeatureKey) -> None:
         # This marks rows as deleted in transaction log without physically removing files
         delta_table.delete()
 
+    def _delete_metadata_impl(
```
danielgafni: needs `current_only` argument

danielgafni (changes requested):
1. A bunch of comments 
2. Looks like hard deletes are not yet (natively) possible with Ibis. I suggest to only keep this PR for soft-deletes. I'll rename it appropriately. Let's either wait with the Ibis hard deletion functionality, or move it to a different PR (we'll just have a separate branch in the stack for hard deletes with Ibis).

danielgafni:
ok, but this PR itself is fine without the ibis hard delete implementation, please re-request review once you address the comments

src/metaxy/metadata_store/base.py
geoHeil: in the past we were returning affected record count; this is still an artifact from that; I think we should remove it now

geoHeil:
comments should be addressed now

src/metaxy/metadata_store/ibis.py
```diff
@@ -453,6 +453,114 @@ def _drop_feature_metadata_impl(self, feature_key: FeatureKey) -> None:
         if table_name in self.conn.list_tables():
             self.conn.drop_table(table_name)
 
+    def _delete_metadata_impl(
```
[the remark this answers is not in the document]
geoHeil: we have NotImplementedError for hard deletion and keep the impl in _delete_metadata_impl for soft deletes

src/metaxy/metadata_store/base.py
```diff
@@ -1597,6 +1617,61 @@ def drop_feature_metadata(self, feature: CoercibleToFeatureKey) -> None:
...
+                [
+                    nw.lit(now).alias(METAXY_DELETED_AT),
+                    (
+                        nw.col(METAXY_CREATED_AT) + nw.lit(timedelta(microseconds=1))
+                    ).alias(METAXY_CREATED_AT),
+                ]
+            )
+            self.write_metadata(feature_key, tombstones.to_native())
+
+        else:
```
danielgafni: add the current feature version filter to this list if `current_only` is set to `True` (currently this is done in backend-specific implementations which is not good)

src/metaxy/metadata_store/ibis.py
```diff
@@ -453,6 +453,99 @@ def _drop_feature_metadata_impl(self, feature_key: FeatureKey) -> None:
...
+        if not filter_list:
+            return
+
+        # Read and filter directly using the store's lazy path
+        filtered = self.read_metadata_in_store(feature_key, filters=filter_list)
+        if filtered is None:
+            return
+
+        # Extract the underlying Ibis table expression with filter applied
+        ibis_filtered = filtered.to_native()
```
danielgafni: pretty sure this is missing a `cast` for the type checker

src/metaxy/metadata_store/ibis.py
```diff
@@ -453,6 +453,99 @@ def _drop_feature_metadata_impl(self, feature_key: FeatureKey) -> None:
...
+                predicate = where_expr.this
+                predicate = predicate.transform(_strip_table_qualifiers())
+                where_clause = predicate.sql()
+                delete_stmt = f"DELETE FROM {table_name} WHERE {where_clause}"
+
+                # Log for debugging
+                import logging
+
+                logger = logging.getLogger(__name__)
+                logger.debug(f"Executing DELETE: {delete_stmt}")
```
danielgafni: can we remove this before mergint the PR?

danielgafni (changes requested):

tests/metadata_stores/test_provenance_golden_reference.py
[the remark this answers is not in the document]
geoHeil: see our previous discussion here https://github.com/anam-org/metaxy/pull/504/changes#r2598561475

src/metaxy/metadata_store/ibis.py
```diff
@@ -453,6 +453,99 @@ def _drop_feature_metadata_impl(self, feature_key: FeatureKey) -> None:
...
+                raise NotImplementedError(
+                    f"Cannot extract WHERE clause for DELETE on {self.__class__.__name__}"
+                )
+        else:
+            raise NotImplementedError(
+                f"Hard delete not yet implemented for {self.__class__.__name__}. "
+                f"Override _get_raw_connection() to provide raw SQL execution capability."
+            )
+
+    def _get_raw_connection(self) -> Any | None:
```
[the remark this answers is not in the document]
geoHeil: done

src/metaxy/metadata_store/base.py
geoHeil: should be done

src/metaxy/metadata_store/ibis.py
```diff
@@ -453,6 +453,99 @@ def _drop_feature_metadata_impl(self, feature_key: FeatureKey) -> None:
...
+                # No WHERE clause - would delete all rows, which is dangerous
+                # Let's print the SQL for debugging
+                import logging
+
+                logger = logging.getLogger(__name__)
+                logger.error(f"Failed to extract WHERE clause from SQL: {select_sql}")
+                raise NotImplementedError(
+                    f"Cannot extract WHERE clause for DELETE on {self.__class__.__name__}"
+                )
+        else:
```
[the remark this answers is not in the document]
geoHeil: should be now using that by default, allowing an override in raw mode

tests/metadata_stores/test_provenance_golden_reference.py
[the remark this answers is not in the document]
danielgafni: Yes, tests that aren’t related to provenance calculations should be moved away from the module that’s testing provenance calculations. 

You can always share the setup via an upper conftest.py, what setup are you talking about?

src/metaxy/metadata_store/ibis.py
```diff
@@ -453,6 +453,79 @@ def _drop_feature_metadata_impl(self, feature_key: FeatureKey) -> None:
...
+        Args:
+            feature_key: Feature to delete from
+            filters: Narwhals expressions to filter records
+        """
+        table_name = self.get_table_name(feature_key)
+
+        current_only = bool(_kwargs.get("current_only", False))
+        if current_only and not self._is_system_table(feature_key):
+            raise NotImplementedError(
+                f"{self.__class__.__name__} does not yet support current_only deletes"
```
geoHeil: for system tables

src/metaxy/metadata_store/ibis.py
```diff
@@ -453,6 +453,76 @@ def _drop_feature_metadata_impl(self, feature_key: FeatureKey) -> None:
         if table_name in self.conn.list_tables():
             self.conn.drop_table(table_name)
 
+    def _delete_metadata_impl(
+        self,
+        feature_key: FeatureKey,
+        filters: Sequence[nw.Expr],
```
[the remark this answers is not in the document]
geoHeil: fixed already

tests/metadata_stores/test_provenance_golden_reference.py
```diff
@@ -64,6 +66,17 @@
 FeaturePlanSequence: TypeAlias = list[FeaturePlanOutput]
 
 
+class UnsupportedDeleteStore(Exception):
```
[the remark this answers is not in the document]
geoHeil: fixed already

src/metaxy/metadata_store/base.py
```diff
@@ -750,17 +756,37 @@ def read_metadata(
...
+                        nw.when(~nw.col(METAXY_DELETED_AT).is_null())
+                        .then(nw.lit("1"))
+                        .otherwise(nw.lit("0"))
+                    )
+                    ordering_expr = nw.concat_str(
+                        [created_at_str, deleted_suffix], separator="|"
+                    )
+                    lazy_frame = lazy_frame.with_columns(
+                        ordering_expr.alias(ordering_column)
+                    )
```
danielgafni: I don't understand this code. 

The `if` branch will always run because the shema always is going to have the METAXY_DELETED_AT column. 

What is the purpose of this ordering column constant if it only appears here? 

I think a much easier fix here is to order by `max(metaxy_created_at, metaxy_deleted_at)`. 

Could you please rewrite this yourself without using an LLM?

src/metaxy/metadata_store/ibis.py
```diff
@@ -453,6 +453,78 @@ def _drop_feature_metadata_impl(self, feature_key: FeatureKey) -> None:
...
+        # Handle empty filters - truncate entire table
+        if not filter_list:
+            if table_name not in self.conn.list_tables():
+                return
+            cast(Any, self.conn).truncate_table(table_name)
+            return
+
+        # Read and filter using store's lazy path to build WHERE clause
+        filtered = self.read_metadata_in_store(feature_key, filters=filter_list)
+        if filtered is None:
```
danielgafni: This probably should be an error as well (FeatureNotFoundError?)

danielgafni (changes requested):
I didn't go through all the tests yet, but testing code needs improvements:

- sharing same data via fixtures. As I understasnd we can mostly reuse the same dataframe in all tests. 
- I found a logical mistake in one test, this may mean more tests may be wrong, and possible the source code ase well? 

Also, the soft deletion read logic doesn't make any sense, please write it yourself or review after LLMs, it's really exhausting to be talking to an LLM all the time

src/metaxy/metadata_store/base.py
```diff
@@ -750,17 +756,37 @@ def read_metadata(
...
                 )
+                schema = lazy_frame.collect_schema()
+                schema_names = schema.names()
+                ordering_column = METAXY_CREATED_AT
+                if METAXY_DELETED_AT in schema_names:
+                    ordering_column = METAXY_ORDERING_TIMESTAMP
+                    created_at_str = nw.col(METAXY_CREATED_AT).cast(nw.String)
+                    deleted_suffix = (
+                        nw.when(~nw.col(METAXY_DELETED_AT).is_null())
+                        .then(nw.lit("1"))
```
geoHeil: this should be a coalesce

src/metaxy/metadata_store/base.py
geoHeil: Problem: delete-by-key vs time revisions

Should we perhaps:

- have a clear, incrementing version number field?
- have a clear definition of what soft delete means? a) delete by key ID fully (all revisions) or b) just the latest one only (that is the more challenging one triggering this edge case)

this code above (complex version)


this one here:

```
ordering_expr = (
                        nw.when(nw.col(METAXY_DELETED_AT).is_null())
                        .then(nw.col(METAXY_CREATED_AT))
                        .otherwise(nw.col(METAXY_DELETED_AT))
                    )
```

(simple version)

Reason:

a delete marker for an older version can end up “winning” over a newer
  active row because its deleted_at is later, which hides the newer active row.

## example

Example timeline (single sample_uid = "a")

```

  t1: write value=1          created_at=t1, deleted_at=NULL
  t2: write value=2          created_at=t2, deleted_at=NULL   (current active)
  t3: soft delete old row    created_at=t1, deleted_at=t3     (delete marker for v1)

```

  If you order by max(created_at, deleted_at):

  - v1’s ordering timestamp = max(t1, t3) = t3
  - v2’s ordering timestamp = max(t2, NULL) = t2
    Since t3 > t2, you would pick v1 (deleted) and hide v2 (active), which is wrong.

  Mermaid sequence diagram

```mermaid
  sequenceDiagram
      participant Store
      participant Sample as sample_uid="a"

      Store->>Sample: write v1 (created_at=t1)
      Store->>Sample: write v2 (created_at=t2)
      Store->>Sample: soft delete v1 (deleted_at=t3)

      Note over Sample: If ordering by max(created_at, deleted_at)
      Note over Sample: v1 orders at t3, v2 orders at t2
      Note over Sample: dedup picks v1 (deleted) -> v2 hidden

```


  - Delete-by-key: you delete “the latest row for key” (e.g. always current row);
    ordering by max(created_at, deleted_at) works.
  - Time revisions: you delete an older revision by created_at (as in tests).
    Ordering by max makes the delete marker look “newer” than the current row and
    can incorrectly hide it.

the tie‑breaker approach (created_at primary, deleted_at only for exact
  ties) preserves both behaviors (but is complex)

geoHeil: (let me -rethink this tomorrow less tired though)

danielgafni: > - delete by key ID fully (all revisions)

If by revisions you mean older records with the same feature version, then yes, a soft deletion should hide all of them. As if they would be deleted physically (soft deletes should effectively act the same as hard deletes for the Metaxy operator for most of use cases).

For a Metaxy operator subsequent writes with the same id_columns should "overwrite" the previous records identiafied by the same ID columns (and feature version of course). At any given point of time the whole set of historical records should only be visivle as the latest, single record. Therefore soft-deletes and hard deletes should also hide/delete historical versions (by default at least).

---

So does using argmax(created, delted) or coalesce work?

src/metaxy/metadata_store/ibis.py
```diff
@@ -453,6 +453,78 @@ def _drop_feature_metadata_impl(self, feature_key: FeatureKey) -> None:
...
+            raise NotImplementedError(
+                f"{self.__class__.__name__} does not support deletes for system tables"
+            )
+
+        table_name = self.get_table_name(feature_key)
+        filter_list = list(filters or [])
+
+        # Handle empty filters - truncate entire table
+        if not filter_list:
+            if table_name not in self.conn.list_tables():
```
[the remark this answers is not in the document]
geoHeil: done

tests/metadata_stores/test_delete_filters.py
```diff
@@ -0,0 +1,142 @@
...
+    )
+
+
+@pytest.mark.parametrize(
+    "case",
+    PREDICATE_CASES,
+    ids=[case.name for case in PREDICATE_CASES],
+)
+@pytest.mark.parametrize("soft", [True, False], ids=["soft", "hard"])
+def test_delete_metadata_accepts_predicate_cases(
```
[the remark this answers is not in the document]
geoHeil: added

geoHeil:
### Merge activity

* **Jan 19, 10:46 PM UTC**: @geoHeil merged this pull request with [Graphite](https://app.graphite.com/github/pr/anam-org/metaxy/504).

PR #505, opened 2025-12-08
hard delete for ibis

geoHeil:


danielgafni (changes requested):
making sure this is blocked for now

src/metaxy/metadata_store/_sql_utils.py
```diff
@@ -0,0 +1,42 @@
+"""SQL helper utilities for MetadataStore backends."""
+
+from __future__ import annotations
+
+import sqlglot
+from sqlglot import exp
+
+
+def predicate_from_select_sql(select_sql: str) -> str:
```
geoHeil: we have now a different predicate generator in the shared utils class; migrate to that one

danielgafni:
Currently, Ibis hard deletions are already implemented in downstack #504. The implementation should be moved here in this case.

danielgafni:
@geoHeil so I guess we don't need this PR since you are bundling hard deletes in #504?

PR #506, opened 2025-12-08
[deltalake] implement hard deletions

geoHeil:


src/metaxy/metadata_store/delta.py
```diff
@@ -322,10 +322,58 @@ def _delete_metadata_impl(
...
+        """
+        from deltalake import DeltaTable
+
+        table_uri = self._feature_uri(feature_key)
+
+        if not self._table_exists(table_uri):
+            return
+
+        # Load Delta table
+        delta_table = DeltaTable(table_uri, storage_options=self.storage_options)
```
danielgafni: we really should have a shared method or property for constructing a delta table - don't we already?

src/metaxy/metadata_store/delta.py
```diff
@@ -322,10 +322,58 @@ def _delete_metadata_impl(
...
+        Returns:
+            SQL WHERE clause predicate string
+
+        """
+        from metaxy.metadata_store.utils import (
+            adapt_trino_query_for_datafusion,
+            narwhals_expr_to_sql_predicate,
+        )
+
+        schema = nw.from_native(
```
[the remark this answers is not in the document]
geoHeil: we have now https://github.com/anam-org/metaxy/pull/626

src/metaxy/metadata_store/delta.py
```diff
@@ -262,6 +260,15 @@ def _cast_enum_to_string(
         """Cast Enum columns to String to avoid delta-rs Utf8View incompatibility."""
         return frame.with_columns(pl.selectors.by_dtype(pl.Enum).cast(pl.Utf8))
 
+    def _open_delta_table(
+        self, table_uri: str, *, without_files: bool = False
```
danielgafni: clearly this can take a `feature: CoercibleToFeatureKey` instead of `table_uri`?

src/metaxy/metadata_store/delta.py
```diff
@@ -345,9 +348,53 @@ def _delete_metadata_impl(
...
+            SQL WHERE clause predicate string
+
+        """
+        from metaxy.metadata_store.utils import narwhals_expr_to_sql_predicate
+
+        schema = self.read_feature_schema_from_store(feature_key)
+        return narwhals_expr_to_sql_predicate(
+            filters,
+            schema,
+            dialect="postgres",
```
danielgafni: `datafusion`!

tests/metadata_stores/test_deletion.py
```diff
@@ -235,7 +235,12 @@ def test_soft_delete_historical_version_preserves_latest(any_store: MetadataStor
 
     _require_supported_delete_store(
         any_store,
-        {"ClickHouseMetadataStore", "DuckDBMetadataStore", "InMemoryMetadataStore"},
+        {
```
danielgafni: I except this to be dropped (requested in #504 to use `@skip_exception` on `NotImplementedError` instead)

src/metaxy/metadata_store/delta.py
geoHeil: sounds correct like it might be the source of the confusion

geoHeil: Lets stick to the PG workaround for now; a more recent version of delta did not fix it. I tested 1.3.2 and raised an issue with datafusion here https://github.com/apache/datafusion/issues/19866

PR #507, opened 2025-12-08
[lance] implement hard deletions

geoHeil:


danielgafni (changes requested):

src/metaxy/metadata_store/lancedb.py
```diff
@@ -317,11 +318,88 @@ def _delete_metadata_impl(
...
+        if not self._table_exists(table_name):
+            raise FeatureNotFoundError(
+                f"Feature {feature_key.to_string()} not found in store; cannot hard delete."
+            )
+
+        table = self._get_table(table_name)
+
+        # Convert Narwhals expressions to SQL WHERE clause using Ibis
+        filter_str = self._narwhals_expr_to_sql(
+            nw.all_horizontal(list(filters), ignore_nulls=False),
```
danielgafni: why is this needed? Lance doesn't accept multiple expressions? add a comment if so

danielgafni (changes requested):

PR #508, opened 2025-12-08
cleanup ibis/clickhouse and use metadata for deletions where available

geoHeil:


danielgafni (changes requested):
I don't think this PR is needed at all?

PR #509, opened 2025-12-08
[cli] add deletion command

geoHeil:


src/metaxy/cli/metadata.py
```diff
@@ -365,6 +374,109 @@ def status(
...
+                hint="Use --filter \"column = 'value'\"",
+            ),
+            format,
+        )
+
+    context = AppContext.get()
+    metadata_store = context.get_store(store)
+
+    with metadata_store:
+        graph = load_graph_for_command(context, None, metadata_store, format)
```
danielgafni: it looks like we actually don't need `graph` here

src/metaxy/cli/metadata.py
```diff
@@ -365,6 +374,109 @@ def status(
...
+            }
+            if errors:
+                output["errors"] = errors
+            print(json.dumps(output, indent=2))
+            if errors:
+                raise SystemExit(1)
+            return
+
+        # plain output
+        data_console.print(f"[bold]Deletion mode:[/bold] {mode}")
```
danielgafni: the `data_console` should be used for `stdout`, I won't think we want to print logs there. 

this should go to `console` instead.

src/metaxy/cli/metadata.py
```diff
@@ -365,6 +367,134 @@ def status(
...
+            name=["--store"],
+            help="Metadata store name (defaults to configured default store).",
+        ),
+    ] = None,
+    filters: FilterArgs | None = None,
+    soft: Annotated[
+        bool,
+        cyclopts.Parameter(
+            name=["--soft"],
+            help="Whether to perform soft delete (append tombstones) vs hard delete (physically remove).",
```
danielgafni: tombstones spotted again

tests/cli/test_cli_metadata.py
```diff
@@ -60,6 +60,31 @@ def _write_sample_metadata(
             SystemTableStorage(store).push_graph_snapshot()
 
 
+def _write_custom_metadata(
```
danielgafni: I think this can be a method on `MetaxyProject`(whhat is the base class - I don't remember)?

danielgafni (changes requested):
Almost LGTM except few small comments

Issue #274, opened 2025-11-12
[integration] vortex


geoHeil:
a bit similar to lance we should also integrate with vortex some day https://docs.vortex.dev/

PR #529, opened 2025-12-10
vortex metadata store

geoHeil:
resolves: https://github.com/anam-org/metaxy/issues/274

## limitations

- no file format just table format
- no remote writes in python api


src/metaxy/metadata_store/vortex.py
```diff
@@ -0,0 +1,315 @@
...
+            layout: Directory layout for feature tables. Options:
+
+                - `"nested"`: Feature tables stored in nested directories `{part1}/{part2}.vortex`
+
+                - `"flat"`: Feature tables stored as `{part1}__{part2}.vortex`
+
+            **kwargs: Forwarded to [metaxy.metadata_store.base.MetadataStore][].
+
+        Raises:
+            NotImplementedError: If a remote path (s3://, gs://, az://) is provided.
```
geoHeil: No remote paths supported in python API

src/metaxy/metadata_store/vortex.py
```diff
@@ -0,0 +1,315 @@
...
+            import vortex
+
+            _ = vortex.open(table_uri)
+            return True
+        except (FileNotFoundError, OSError, Exception):
+            return False
+
+    # ===== Storage operations =====
+
+    def write_metadata_to_store(
```
geoHeil: just a file format - not a table format like ducklake, iceberg, delta. This means some extra complexity

geoHeil: https://spiceai.org/blog/releases/v1.9.0 Cayenne Data Accelerator  could be interesting here

geoHeil: and https://github.com/spiraldb/iceberg something in JVM land

PR #551, opened 2025-12-12
[core] fix store metadata with fallback stores

### LINKED ISSUE #549: [core] `MetadataStore.get_store_metadata` should respect fallback stores
Currently it's always returning metadata from the current store even if the feature is only found in a fallback store

danielgafni:
Resolve #549


danielgafni:
### Merge activity

* **Dec 12, 1:03 AM UTC**: @danielgafni merged this pull request with [Graphite](https://app.graphite.com/github/pr/anam-org/metaxy/551).

Issue #556, opened 2025-12-12
[core] Support join types other than `inner`


danielgafni:
Currently we always do `inner` joins when building feature inputs from upstream metadata. 

We should also support `left` joins for "optional" dependencies (at sample level).  

This should be a new parameter on `FeatureDep`. 

A section on optional dependencies has to be added to the docs. 

---

Not sure but we need to check which is the "left most" table we use when joining upstream metadata. Not sure how to handle this correctly. Should there be some kind of a "main" feature we start from when buildings inputs? Should we use `outer` joins instead?

PR #558, opened 2025-12-12
[core] support optional dependencies

geoHeil:
resolves: https://github.com/anam-org/metaxy/issues/556

### DIFF
--- a/docs/guide/learn/optional-dependencies.md
+++ b/docs/guide/learn/optional-dependencies.md
@@ -1,0 +1,91 @@
+# Optional Dependencies
+
+By default, feature dependencies use **inner joins** - a sample only exists in the downstream feature if it exists in **all** upstream dependencies.
+**Optional dependencies** allow you to preserve samples even when some upstream data is missing by using **left joins** instead.
+
+## Overview
+
+When building features with multiple dependencies, you often want certain dependencies to be "optional" - meaning the downstream feature should still include samples even if the optional dependency has no matching data.
+This is controlled by the `optional` parameter on [`FeatureDep`][metaxy.FeatureDep].
+
+```python
+from metaxy import FeatureSpec, FeatureDep
+
+spec = FeatureSpec(
+    key="enriched/video",
+    id_columns=["video_id"],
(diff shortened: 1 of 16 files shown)


danielgafni:
I've had a second thought on this: I think you are right and we don't need outer joins, we just need to start joining from required features, do `inner` joins, then start doing `left` joins once we get to non-required features.

docs/guide/learn/optional-dependencies.md
```diff
@@ -0,0 +1,253 @@
...
+    id_columns=["id"],
+    deps=[
+        FeatureDep(feature=RequiredData),  # Required (first)
+        FeatureDep(feature=OptionalData, optional=True),  # Optional (subsequent)
+    ],
+    fields=["output"],
+)
+```
+
+### Join Order Follows Declaration Order
```
danielgafni: I don't think this matters given we remove this requirement.

Dependency specification ordering should only affect provenance calculations, not actual data. This is just because provenance calculations are ordered in nature.

src/metaxy/models/feature_spec.py
```diff
@@ -302,6 +319,22 @@ def validate_id_columns(self) -> Self:
             )
         return self
 
+    @pydantic.model_validator(mode="after")
+    def validate_first_dep_not_optional(self) -> Self:
```
danielgafni: this should be removed

src/metaxy/models/plan.py
```diff
@@ -345,3 +345,24 @@ def input_id_columns(self) -> list[str]:
...
+        """Dependencies marked as optional (use left join)."""
+        return [dep for dep in (self.feature_deps or []) if dep.optional]
+
+    @cached_property
+    def required_deps(self) -> list[FeatureDep]:
+        """Dependencies that are required (use inner join)."""
+        return [dep for dep in (self.feature_deps or []) if not dep.optional]
+
+    @cached_property
+    def main_dep(self) -> FeatureDep | None:
```
danielgafni: this should be removed

Issue #582, opened 2026-01-02
[deltalake] use `Deltatable.is_deltalake` for feature existence check and use `deltalake > 1.3.0` for tests


geoHeil:
Original `delta-rs` [issue](https://github.com/delta-io/delta-rs/issues/3942#issuecomment-3706087225)

danielgafni:
We should only attempt it in our `dev` dependencies, not the actual package dependencies. Reasons:

1. The current version works perfectly fine outside of our testing environment
2. [This](https://github.com/delta-io/delta-rs/issues/3942#issuecomment-3706087225) comment suggests the original issue might not have been actually fixed? We'd also have to remove our current `try: except` work-around to `.is_deltatable()` call in order to verify the fix is working as expected. 

I think the right approach here is to: 

1. Pin `deltalake >= 1.3.0` in our `dev` dependency group
2. use `.is_deltalake` for `.has_feature` check

geoHeil:
would halt this for now https://github.com/anam-org/metaxy/pull/585#issuecomment-3706878079 also seems to not work for us

PR #583, opened 2026-01-02
upgrade delta constraint and fix ty

geoHeil:
resolves: https://github.com/anam-org/metaxy/issues/582

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,7 +57,7 @@ duckdb = [
   "duckdb>=1.4.1",
 ]
 delta = [
-  "deltalake>=1.2.1",
+  "deltalake>=1.3.0",
 ]
 dagster = [
   "dagster>=1.12.3",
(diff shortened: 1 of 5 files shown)


geoHeil:
I think we do not have to rush this one; should we split it into 2x perhaps 1) for the new version of delta and 2) one for the uv upgrade which incluees and upgrade for ty? (and the hotfixes)

danielgafni:
Honestly, I'm not sure we should enforce the `>1.3.0` deltalake pin in our requirements. 

The deltalake issue has only affected our testing environment (`moto` S3 server specifically), the current constraint works perfectly fine with Metaxy. 

We shouldn't force this upgrade on our users since Metaxy itself doesn't require it at runtime. 

---

The `ty` thing is ok to fix. Upgrading it would be nice as well. But how are these Dagster integration changes related to `ty`?

PR #585, opened 2026-01-03
upgrade deltalake to fix workaround

geoHeil:
resolves: https://github.com/anam-org/metaxy/issues/582


geoHeil:
@danielgafni I fear this does not yet resolve the problem - locally the test now fails if we remove the workaround

geoHeil:
if CI errors show the same I would suggest we do NOT perform the delta upgrade/removal of the fix for now

danielgafni (commented):
So this means their fix didn’t solve the problem for us. The issue in delta-rs should be reopened.

PR #605, opened 2026-01-13
[core] Add SQL generation utilities for Narwhals expressions

geoHeil:


src/metaxy/metadata_store/utils.py
```diff
@@ -102,3 +109,214 @@ def sanitize_uri(uri: str) -> str:
...
+                    return schema_attr()
+                return ibis.schema(schema_attr)
+            if hasattr(source, "to_arrow"):
+                return ibis.schema(source.to_arrow().schema)
+            raise
+
+    ibis_table = ibis.table(_infer_ibis_schema(schema_source), name="temp_table")
+    nw_lf = nw.from_native(ibis_table, eager_only=False)
+    result_lf = narwhals_function(nw_lf)
+    ibis_expr = result_lf.to_native()
```
danielgafni: Not sure but I think we can just `return result_ldf.sql(dialect)`?

src/metaxy/metadata_store/utils.py
```diff
@@ -102,3 +109,214 @@ def sanitize_uri(uri: str) -> str:
         pass
 
     return uri
+
+
+def convert_make_timestamptz_to_cast(sql: str, *, use_arrow_cast: bool = False) -> str:
```
danielgafni: Honestly this function looks super weird, there must be an easier solution

danielgafni (changes requested):

src/metaxy/metadata_store/utils.py
```diff
@@ -102,3 +109,214 @@ def sanitize_uri(uri: str) -> str:
...
+            raise
+
+    ibis_table = ibis.table(_infer_ibis_schema(schema_source), name="temp_table")
+    nw_lf = nw.from_native(ibis_table, eager_only=False)
+    result_lf = narwhals_function(nw_lf)
+    ibis_expr = result_lf.to_native()
+    return ibis.to_sql(ibis_expr, dialect=dialect)
+
+
+def narwhals_expr_to_sql_predicate(
```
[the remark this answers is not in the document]
geoHeil: it was the difference between what lance and what delta needed; it at least happend before for delta the temporary table aliases that were generated from our existing code paths so I would assume we cannot remove it. Yes we can set it to false for lance.

src/metaxy/metadata_store/utils.py
geoHeil: we anywys only need it for lance

geoHeil: cleaned it up wiht your recommendation

src/metaxy/metadata_store/utils.py
geoHeil: seems to work fine - moved to this simpler approach

src/metaxy/metadata_store/utils.py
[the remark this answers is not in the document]
geoHeil: refactored

src/metaxy/metadata_store/utils.py
```diff
@@ -102,3 +113,192 @@ def sanitize_uri(uri: str) -> str:
...
+    )
+    sql = generate_sql(lambda lf: lf.filter(*filter_list), schema, dialect=dialect)
+
+    import sqlglot
+    from sqlglot import exp
+
+    parsed = sqlglot.parse_one(sql, read=dialect)
+    where_expr = parsed.args.get("where")
+    if not where_expr:
+        raise NotImplementedError(
```
danielgafni: RuntimeError?

src/metaxy/metadata_store/utils.py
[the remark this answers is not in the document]
danielgafni: I see. But is it even a valid action to do? Are these aliases always redundant? Do you think this would work if we do latest_only=True during deletion? The query would then involve a groupby, sort by timestamp, take latest only. Not that we have to support it, I’m just curious. We can start with only supporting simple, flat WHERE. 



Another question: have you tried using sqlframe instead? I’d guess it might produce different queries.

src/metaxy/metadata_store/utils.py
```diff
@@ -102,3 +113,192 @@ def sanitize_uri(uri: str) -> str:
...
+    ibis_table = ibis.table(_to_ibis_schema(schema), name="temp_table")
+    nw_lf = nw.from_native(ibis_table, eager_only=False)
+    result_lf = narwhals_function(nw_lf)
+    ibis_expr = result_lf.to_native()
+    return ibis.to_sql(ibis_expr, dialect=dialect)
+
+
+def narwhals_expr_to_sql_predicate(
+    filters: nw.Expr | Sequence[nw.Expr],
+    schema: NarwhalsSchema | FrameT,
```
danielgafni: Can be just Frame instead of FrameT since it’s not repeated and we don’t need it to be a TypeVar

src/metaxy/metadata_store/utils.py
danielgafni: But have you tried getting rid of it and using the Trino dialect for SQL generation - perhaps LanceDB and DeltaLake would just accept it right away?

src/metaxy/metadata_store/utils.py
```diff
@@ -102,3 +114,117 @@ def sanitize_uri(uri: str) -> str:
         pass
 
     return uri
+
+
+def generate_sql(
+    narwhals_function: Callable[[nw.LazyFrame[Any]], nw.LazyFrame[Any]],
```
danielgafni: this can have `Frame` inside the `Callable`

src/metaxy/metadata_store/utils.py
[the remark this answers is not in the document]
geoHeil: so far not sqlframe; lets keep a simple lfat where for now; I further simplified/deleted stuff and wil test it soon with restacked lance/delta and hope it works in a simpler approach

src/metaxy/metadata_store/utils.py
```diff
@@ -1,9 +1,21 @@
...
 
-from narwhals.typing import FrameT
+import narwhals as nw
+from narwhals.typing import Frame, FrameT
+from sqlglot import exp
+
+if TYPE_CHECKING:
+    from collections.abc import Callable
+
+    from narwhals.typing import Schema as NarwhalsSchema
```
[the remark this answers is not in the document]
geoHeil: why? isnt that a type checking only thing?

src/metaxy/metadata_store/utils.py
```diff
@@ -102,3 +112,98 @@ def sanitize_uri(uri: str) -> str:
...
+    result_lf = narwhals_function(nw_lf)
+    ibis_expr = result_lf.to_native()
+    return ibis.to_sql(ibis_expr, dialect=dialect)
+
+
+def narwhals_expr_to_sql_predicate(
+    filters: nw.Expr | Sequence[nw.Expr],
+    schema: NarwhalsSchema | Frame,
+    *,
+    dialect: str = "duckdb",
```
danielgafni: can we remove the default value here? might cause unexpected errors for callers

src/metaxy/metadata_store/utils.py
```diff
@@ -102,3 +112,98 @@ def sanitize_uri(uri: str) -> str:
...
+    sql = generate_sql(lambda lf: lf.filter(*filter_list), schema, dialect=dialect)
+
+    import sqlglot
+    from sqlglot import exp
+    from sqlglot.optimizer.simplify import simplify
+
+    parsed = sqlglot.parse_one(sql, read=dialect)
+    where_expr = parsed.args.get("where")
+    if not where_expr:
+        return exp.true().sql(dialect=dialect)
```
danielgafni: I feel like this may lead to unexpected results... 

Can we instead just raise an error here?

src/metaxy/metadata_store/utils.py
```diff
@@ -102,3 +112,98 @@ def sanitize_uri(uri: str) -> str:
...
+    ibis_table = ibis.table(_to_ibis_schema(schema), name="_metaxy_temp")
+    nw_lf = nw.from_native(ibis_table, eager_only=False)
+    result_lf = narwhals_function(nw_lf)
+    ibis_expr = result_lf.to_native()
+    return ibis.to_sql(ibis_expr, dialect=dialect)
+
+
+def narwhals_expr_to_sql_predicate(
+    filters: nw.Expr | Sequence[nw.Expr],
+    schema: NarwhalsSchema | Frame,
```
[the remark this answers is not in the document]
geoHeil: removed

src/metaxy/metadata_store/utils.py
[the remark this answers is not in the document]
geoHeil: the docs build then fails 

```
from narwhals.typing import Schema as NarwhalsSchema
  ImportError: cannot import name 'Schema' from 'narwhals.typing' (/Users/geoheil/development/metaxy/metaxy/.venv/lib/python3.10/site-packages/narwhals/typing.py)
```

so for now I would actually suggeset we keep it as is if that is fine for you

tests/metadata_stores/test_utils.py
```diff
@@ -18,3 +33,104 @@ def test_is_local_path() -> None:
...
+    parsed = parse_one(sql, read="duckdb")
+    where_expr = parsed.args.get("where")
+    assert where_expr is not None
+    assert isinstance(where_expr.this, exp.GT)
+    column = where_expr.this.find(exp.Column)
+    assert column is not None
+    assert column.name == "price"
+
+
+def test_narwhals_expr_to_sql_predicate_no_filters_returns_true() -> None:
```
danielgafni: this should raise an error

tests/metadata_stores/test_utils.py
```diff
@@ -18,3 +33,104 @@ def test_is_local_path() -> None:
...
+    schema = nw.from_native(
+        pl.DataFrame(schema={"date": pl.Date, "price": pl.Float64})
+    ).collect_schema()
+    predicate = narwhals_expr_to_sql_predicate(
+        nw.col("price") > 5, schema, dialect="duckdb"
+    )
+    assert "_metaxy_temp" not in predicate
+
+
+def test_narwhals_expr_to_sql_predicate_simplifies() -> None:
```
danielgafni: I don't think this test asserts anything useful

src/metaxy/metadata_store/utils.py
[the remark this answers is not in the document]
danielgafni: I just checked - indeed, it's because it's not exported in `narwhals.typing`: it's hidden behind `if TYPE_CHECKING` there (for reasons unknown, I guess it's just an alias for cases where users don't need narwhals at runtime).

It should be either imported as `from narwhals import Schema` (at runtime) or just use as `nw.Schema` - I don't think we even need the alias.

All of this isn't important, but perhaps it clarifies things for you.

src/metaxy/metadata_store/utils.py
```diff
@@ -102,3 +114,164 @@ def sanitize_uri(uri: str) -> str:
         pass
 
     return uri
+
+
+def convert_make_timestamptz_to_cast(
```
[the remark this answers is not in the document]
danielgafni: ok that's fine

danielgafni (changes requested):
Metaxy changes seem good now, but testing is clearly lacking. 

**This is critical functionality which is going to be used for permanent deletes**. We cannot leave it untested.

Please add more tests for SQL generation and more tests for our datafusion shim. 

Apart from logical tests, it's easy to add property-based testing with hypothesis: we can run SQL generated from the same sets of filter expressions through duckdb and datafusion and make sure they evaluate to the same records. This would be similar to the golden test suite that we have for `resolve_update`. It's very easy to do over a Parquet file or a Polars DataFrame or smth. This should be a single test parametrized over `filters` argument. Additionally, each time the result can be snapshotted via `syrupy` to detect unwanted changes in the future.

src/metaxy/metadata_store/utils.py
```diff
@@ -102,3 +114,164 @@ def sanitize_uri(uri: str) -> str:
...
+        >>> filters = nw.col("status") == "active"
+        >>> narwhals_expr_to_sql_predicate(filters, df, dialect="duckdb")
+        '"status" = \'active\''
+    """
+    filter_list = (
+        list(filters)
+        if isinstance(filters, Sequence) and not isinstance(filters, nw.Expr)
+        else [filters]
+    )
+    if not filter_list:
```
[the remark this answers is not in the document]
geoHeil: removed, raising now `ValueError`

src/metaxy/metadata_store/utils.py
```diff
@@ -102,3 +112,172 @@ def sanitize_uri(uri: str) -> str:
...
+
+            return exp.Cast(
+                this=exp.Literal.string(iso_str),
+                to=exp.DataType.build("TIMESTAMP"),
+            )
+
+        return node
+
+    transformed = parsed.transform(_adapt)
+    return transformed.sql(dialect="trino")
```
danielgafni: oh so `.sql()` without dialect would just keep it as is? 

perhaps that's what we should do here since the result can already diverge from `trino` at this point?

danielgafni (changes requested):

src/metaxy/metadata_store/utils.py
geoHeil: it does not seem to matter. (from the tests) it is shorter without, more explicit with. Do you have a preference?

tests/metadata_stores/test_utils.py
```diff
@@ -18,3 +36,196 @@ def test_is_local_path() -> None:
...
+
+    ctx = datafusion.SessionContext()
+    ctx.register_record_batches("t", [table.to_batches()])
+    df = ctx.sql(f"SELECT id FROM t WHERE {datafusion_predicate} ORDER BY id")
+    batches = df.collect()
+    datafusion_ids: list[int] = []
+    for batch in batches:
+        datafusion_ids.extend(batch.column(0).to_pylist())
+
+    assert duckdb_ids == datafusion_ids
```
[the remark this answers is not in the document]
geoHeil: done

tests/metadata_stores/test_utils.py
```diff
@@ -18,3 +35,222 @@ def test_is_local_path() -> None:
...
+    "case_fn",
+    [
+        _predicate_case_basic,
+        _predicate_case_tz_timestamp,
+        _predicate_case_naive_timestamp,
+        _predicate_case_is_in,
+    ],
+    ids=["basic", "tz_timestamp", "naive_timestamp", "is_in"],
+)
+def test_narwhals_expr_to_sql_predicate_outputs(
```
danielgafni: if you want to have a test that only snapshots it make sense to have `dialect` as an parameter for the test, not inside the dict

danielgafni (changes requested):
We also need a test that would ensure that a more complicated expression (such as a window function) either works with the current code or raises an excewption. 

Otherwise we may generate a wrong query and delete wrong data for the users. 

This would be bad. 

I guess we also need to add a warning about this somewhere in the docs.

tests/metadata_stores/test_utils.py
```diff
@@ -18,3 +36,196 @@ def test_is_local_path() -> None:
...
+        elif op == "<=":
+            exprs.append(col <= val)
+        elif op == ">":
+            exprs.append(col > val)
+        elif op == ">=":
+            exprs.append(col >= val)
+        elif op == "==":
+            exprs.append(col == val)
+        else:
+            exprs.append(col != val)
```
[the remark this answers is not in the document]
geoHeil: removed now

src/metaxy/metadata_store/utils.py
```diff
@@ -102,3 +112,94 @@ def sanitize_uri(uri: str) -> str:
...
+        schema: Narwhals schema to build the Ibis table
+
+    Returns:
+        SQL WHERE clause predicate string (without the "WHERE" keyword)
+
+    Raises:
+        RuntimeError: If WHERE clause cannot be extracted from generated SQL
+
+    Example:
+        >>> import narwhals as nw
```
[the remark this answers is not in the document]
geoHeil: fixed

tests/metadata_stores/test_utils.py
geoHeil: you mean like we have it now for parametrize?

tests/metadata_stores/test_utils.py
```diff
@@ -18,3 +33,313 @@ def test_is_local_path() -> None:
...
+            ],
+        ),
+    ],
+)
+def test_predicate_consistency_duckdb_datafusion_snapshot(
+    case_name: str,
+    exprs: list[nw.Expr],
+    snapshot: SnapshotAssertion,
+) -> None:
+    duckdb = pytest.importorskip("duckdb")
```
danielgafni: we don't do this atm, I don't want LLMs to pick this up

docs/guide/learn/metadata-stores.md
```diff
@@ -30,6 +30,12 @@ Deletes are not required during normal operations, but they are still supported
 
     1. :fire: for performance reasons
 
+!!! warning
+
+    Delete filters must compile to SQL WHERE clauses. Expressions that rely on window functions or QUALIFY-style
+    predicates are not supported for deletions and will raise errors. Keep delete filters simple to avoid
```
danielgafni: > Keep delete filters simple to avoid accidental data loss.

Maybe let's avoid this kind of language? This is looking too scary and unreliable. I think the first part is enough.

PR #624, opened 2026-01-15
add sqlglot dialect to IBIS metadata stores

geoHeil:


danielgafni (changes requested):
Can you first check if the dialect can be extracted from the available Ibis backend? It seems like that's the case. From Ibis [docs](https://ibis-project.org/backends/datafusion#ibis.backends.datafusion.Backend.sql): 

```
The SQL dialect of the query. If not provided, the backend’s dialect is assumed. This argument can be useful when the query is written in a different dialect from the backend.
```
Does the backend name just match the dialect name?

danielgafni:
You can just return `self._conn.dialect` (found an example usage [here](https://github.com/ibis-project/ibis/blob/main/ibis/backends/datafusion/__init__.py#L308)) from a `@property` `dialect` on `MetadataStore`

--

P.S. not sure if `_conn` is the Ibis `Backend` object or it's something else

PR #626, opened 2026-01-15
[core] read_store_schema

geoHeil:
[core] read_store_schema

### DIFF
--- a/src/metaxy/metadata_store/base.py
+++ b/src/metaxy/metadata_store/base.py
@@ -1620,6 +1620,27 @@ class MetadataStore(ABC):
         """
         pass
 
+    def read_feature_schema_from_store(
+        self,
+        feature: CoercibleToFeatureKey,
+    ) -> nw.Schema:
+        """Read the schema for a feature from the store.
+
+        Args:
+            feature: Feature to read schema for
+
+        Returns:
+            Narwhals schema for the feature
+
+        Raises:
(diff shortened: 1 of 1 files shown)


src/metaxy/metadata_store/base.py
```diff
@@ -1703,6 +1703,33 @@ def read_metadata_in_store(
         """
         pass
 
+    def read_store_schema(
+        self,
+        feature: CoercibleToFeatureKey,
+        *,
+        allow_fallback: bool = False,
```
danielgafni: I feel like this argument is not needed yet: 

1. We currently only need this method for hard deletes, and they do not use fallback stores 
2. It's a bit weird to have it default to `False` since in onther cases we typically default to using fallback stores 

I think this argument can have it's place in a different method that would be called something like `read_schema` (without emphasis on using this specific store). But we don't need it right now.

src/metaxy/metadata_store/base.py
```diff
@@ -1620,6 +1620,27 @@ def read_metadata_in_store(
...
+        """Read the schema for a feature from the store.
+
+        Args:
+            feature: Feature to read schema for
+
+        Returns:
+            Narwhals schema for the feature
+
+        Raises:
+            FeatureNotFoundError: If feature not found in any store
```
danielgafni: "not found in the store"

danielgafni (changes requested):

PR #652, opened 2026-01-16
[examples] fix uv misconfiguration

danielgafni:
This basically hijacks `uv` workspace - I am not sure this setup is going to stay legal forever. 

This is because from the root package perspective, `examples/example-*` are all `uv` workspace members, and `examples` is a workspace member as well. For some reason this doesn't immediately trigger nested workspace detection (not supported by `uv`). 

Meanwhile, `examples/example-*` doesn't consider itself a root workspace member. Instead, `metaxy` is added there as an editable path dependency. 

This may stop working in future `uv` versions, but should work for now until we fix this properly.


danielgafni:
* **#652** <a href="https://app.graphite.com/github/pr/anam-org/metaxy/652?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anam-org/metaxy/652?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #760, opened 2026-01-23
[ray] add datasink

danielgafni:


danielgafni:
> If i'm not mistaken, mx.init_metaxy(self.config) feels like it could get shunted to on_write_start

Actually, Metaxy has to be initialized in the Ray worker, and it seems like `on_write_start` runs only on the driver.

Issue #443, opened 2025-12-01
[FeatureDefinition] Push `FeatureDefinition` to the system features table


danielgafni:
Maybe rename it to `feature_definitions`

PR #795, opened 2026-01-27
[core] use FeatureDefinition everywhere

### LINKED ISSUE #437: [FeatureDefinition] Use `FeatureDefinition` in `FeatureGraph`
Use `FeatureDefinition` in `FeatureGraph`, avoid using feature classes whatsoever. `FeatureGraph` should **only contain `FeatureDefinition`** objects.

danielgafni:
Resolve #437 #443


src/metaxy/ext/dagster/table_metadata.py
```diff
@@ -20,32 +20,115 @@
 )
 
 
-def build_column_schema(feature_cls: type[mx.BaseFeature]) -> dg.TableSchema:
-    """Build a Dagster TableSchema from a Metaxy feature class.
+def _get_type_string_from_schema(field_props: dict[str, Any]) -> str:
```
danielgafni: I don't like the way this is implemented

It is very easy to get it wrong. 

Is there a native, official way to parse Pydantic fields back from the serialized schema instead?

src/metaxy/models/feature.py
```diff
@@ -119,26 +83,38 @@ def add_feature(self, feature: type["BaseFeature"]) -> None:
         if feature.spec().deps:
             self._build_and_validate_plan(feature.spec())
 
-        self.features_by_key[feature.spec().key] = feature
-        self.feature_specs_by_key[feature.spec().key] = feature.spec()
+        # Create FeatureDefinition from the class
+        definition = FeatureDefinition.from_feature_class(feature)
+
+        # Store the definition
+        self.feature_definitions_by_key[key] = definition
```
danielgafni: I think we should call `self.add_feature_definition` instead? it can do additional validation

src/metaxy/models/feature.py
```diff
@@ -600,136 +577,50 @@ class VideoProcessing(mx.BaseFeature, spec=mx.FeatureSpec(key="video_processing"
...
+
+        for feature_key_str, feature_data in snapshot_data.items():
+            # Parse FeatureSpec
+            feature_spec_dict = feature_data["feature_spec"]
+            spec = FeatureSpec.model_validate(feature_spec_dict)
+
+            # Get other fields with defaults for backward compatibility
+            feature_schema = feature_data.get("feature_schema", {})
+            feature_class_path = feature_data.get("feature_class_path", "")
+            project = feature_data.get("project", spec.key.parts[0] if spec.key.parts else "")
```
danielgafni: we should not assign empty defaults, that's bad data

PR #955, opened 2026-02-16
Fixing ducklake

geoHeil:
Fixing ducklake

### DIFF
--- a/docs/examples/ducklake.md
+++ b/docs/examples/ducklake.md
@@ -1,0 +1,67 @@
+---
+title: "DuckLake Example"
+description: "Example of configuring DuckLake as an open lakehouse format for the DuckDB metadata store."
+---
+
+# DuckLake
+
+## Overview
+
+::: metaxy-example source-link
+    example: ducklake
+
+This example demonstrates how to configure [DuckLake](https://ducklake.select/) with the DuckDB metadata store.
+DuckLake is an open lakehouse format that separates the metadata catalog (table definitions, schema evolution, and transaction history) from data file storage.
+This lets you choose independent backends for each layer, for example PostgreSQL for the catalog and S3 for data files.
+
+We will set up a DuckLake-backed store via `metaxy.toml` and preview the SQL statements that DuckLake would execute when attaching to a DuckDB connection.
+
+## Getting Started
+
+Install the example's dependencies:
+
+```shell
+uv sync
+```
+
+## Configuration
+
+DuckLake is configured in `metaxy.toml` with two parts: a **metadata backend** (where the catalog is stored) and a **storage backend** (where data files live).
+
+The active configuration below uses SQLite for the metadata catalog and the local filesystem for data storage. Commented-out sections show alternative backends.
+
+```toml title="metaxy.toml"
+--8<-- "example-ducklake/metaxy.toml"
+```
+
+Available backend combinations:
+
+| Metadata backend | Storage backend |
+|---|---|
+| DuckDB, SQLite, PostgreSQL | local filesystem, S3, Cloudflare R2, Google Cloud Storage |
+| MotherDuck | managed (no storage backend needed) |
+
+!!! tip
+
+    Omit `key_id` and `secret` from S3, R2, or GCS storage backends to use the credential chain (IAM roles, environment variables, etc.) instead of static credentials.
+
+## Walkthrough
+
+The demo script initializes the store from configuration and previews the SQL statements that would be executed:
+
+```python title="src/example_ducklake/demo.py"
+--8<-- "example-ducklake/src/example_ducklake/demo.py"
+```
+
+Run the demo:
+
+```shell
+uv run python src/example_ducklake/demo.py
+```
+
+The output shows the full sequence of SQL statements: creating secrets for the metadata and storage backends, and attaching the DuckLake database.
+
+## Related Materials
+
+- [DuckDB Metadata Store](../integrations/metadata-stores/databases/duckdb.md)
+- [`DuckLakeAttachmentConfig`][metaxy.ext.metadata_stores._ducklake_support.DuckLakeAttachmentConfig]
--- a/src/metaxy/ext/metadata_stores/_ducklake_support.py
+++ b/src/metaxy/ext/metadata_stores/_ducklake_support.py
@@ -1,174 +1,227 @@
 """Shared DuckLake configuration helpers."""
 
-import os
-from collections.abc import Mapping, Sequence
-from typing import Any, Protocol, runtime_checkable
+from collections.abc import Mapping
+from typing import Annotated, Any, Literal
 
 from duckdb import DuckDBPyConnection  # noqa: TID252
 from pydantic import (
     BaseModel,
-    ConfigDict,
     Field,
-    ValidationInfo,
-    computed_field,
     field_validator,
+    model_validator,
 )
+from typing_extensions import Self
 
 from metaxy._decorators import public
 
-
-@runtime_checkable
-class SupportsDuckLakeParts(Protocol):
-    """Protocol for objects that can produce DuckLake attachment SQL fragments."""
-
-    def get_ducklake_sql_parts(self, alias: str) -> tuple[str, str]: ...
+# ---------------------------------------------------------------------------
+# Metadata backend configs
+# ---------------------------------------------------------------------------
 
 
-@runtime_checkable
-class SupportsModelDump(Protocol):
-    """Protocol for Pydantic-like objects that expose a model_dump method."""
+class DuckDBMetadataBackendConfig(BaseModel):
+    """DuckDB file-based metadata backend for [DuckLake](https://ducklake.select/)."""
 
-    def model_dump(self) -> Mapping[str, Any]: ...
+    type: Literal["duckdb"] = "duckdb"
+    uri: str
 
-
-DuckLakeBackendInput = Mapping[str, Any] | SupportsDuckLakeParts | SupportsModelDump
-DuckLakeBackend = SupportsDuckLakeParts | dict[str, Any]
-
-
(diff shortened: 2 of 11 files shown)


src/metaxy/ext/metadata_stores/_ducklake_support.py
danielgafni: I think the code would be safer if we removed this `ConfigDict` - why would we need extra fields? arbitrary types better not go here as well, you only need it right now because `DuckLakeBackend` accepts an actual DuckDB connetion object, but this should not be the case.

danielgafni: a `description` would be nice as well

src/metaxy/ext/metadata_stores/duckdb.py
danielgafni: what does this mean?

danielgafni: not sure if this is good, what other fields would we need here?

danielgafni: This should be typed with the exact coerced final type instead of a type union.

src/metaxy/ext/metadata_stores/_ducklake_support.py
danielgafni: Why do we need `_PreviewConnection`? 
Is this only for the "preview" functionality? Why is it needed? Is it only for testing? Then the tests should take care of it.

src/metaxy/ext/metadata_stores/_ducklake_support.py
```diff
@@ -2,173 +2,161 @@
...
+            for key, value in self.secret_parameters.items():
+                secret_params[str(key).upper()] = value
+
+        secret_name = f"secret_catalog_{alias}"
+        secret_sql = build_secret_sql(secret_name, "postgres", secret_params)
+        metadata_params = f"METADATA_PATH '', METADATA_PARAMETERS MAP {{'TYPE': 'postgres', 'SECRET': '{secret_name}'}}"
+        return secret_sql, metadata_params
+
+
+DuckLakeMetadataBackend = Annotated[
```
danielgafni: maybe inline it in the consumer class if we don't use it anywhere else

src/metaxy/ext/metadata_stores/_ducklake_support.py
```diff
@@ -2,173 +2,161 @@
...
+
+    type: Literal["s3"]
+    secret: Mapping[str, Any] | None = None
+    aws_access_key_id: str | None = None
+    aws_secret_access_key: str | None = None
+    endpoint_url: str | None = None
+    bucket: str | None = None
+    prefix: str | None = None
+    url_style: str = "path"
+    region: str = "us-east-1"
```
danielgafni: let's remove the default

src/metaxy/ext/metadata_stores/_ducklake_support.py
```diff
@@ -303,48 +266,60 @@ class DuckLakeAttachmentManager:
...
-        preview_conn = _PreviewConnection()
-        self.configure(preview_conn)
-        return preview_conn.cursor().commands
-
-
-DuckLakeConfigInput = DuckLakeAttachmentConfig | Mapping[str, Any]
+        return self._build_sql_statements()
 
 
 def build_ducklake_attachment(
```
danielgafni: looks like we don't need it anymore since we can always work with `DuckLakeAttachmentConfig`

src/metaxy/ext/metadata_stores/duckdb.py
```diff
@@ -266,28 +246,22 @@ def _load_extensions(self) -> None:
...
+            row[0]
+            for row in duckdb_conn.execute("SELECT extension_name FROM duckdb_extensions() WHERE installed").fetchall()
+        }
+
+        for ext in self.extensions:
+            if ext.name not in installed:
+                # Default to community repository; the Python API defaults to
+                # the core repository which doesn't host community extensions.
+                duckdb_conn.install_extension(ext.name, repository=ext.repository or "community")
+            duckdb_conn.load_extension(ext.name)
```
danielgafni: Let's just loop over extensions and install each. 
We can keep track of installed extensions in a MetadataStore attribute.

tests/metadata_stores/conftest.py
```diff
@@ -99,6 +99,20 @@ def case_duckdb(self, tmp_path: Path) -> MetadataStore:
             extensions=["hashfuncs"],
         )
 
+    @pytest.mark.ibis
+    @pytest.mark.native
+    @pytest.mark.duckdb
+    def case_duckdb_ducklake(self, tmp_path: Path) -> MetadataStore:
```
danielgafni: let's add a `@pytest.mark.ducklake`

danielgafni (changes requested):

tests/metadata_stores/conftest.py
geoHeil: you mean both duckdb + ducklake on this one? or globally add and then here only have ducklake? (as we have now)

src/metaxy/ext/metadata_stores/duckdb.py
geoHeil: done

src/metaxy/ext/metadata_stores/_ducklake_support.py
```diff
@@ -1,174 +1,219 @@
...
+
+    def sql_parts(self, alias: str) -> tuple[str, str]:
+        return "", f"DATA_PATH {_stringify_scalar(self.path)}"
+
+
+class S3AuthConfig(BaseModel):
+    """Explicit S3 credentials for DuckLake storage."""
+
+    aws_access_key_id: str
+    aws_secret_access_key: str
```
danielgafni: I'd say separating just these two params is not worth it 

Maybe let's flatten it back into `S3StorageBackendConfig`

Also, let's rename them to `access_key_id` and `secret_access_key`, but add Pydantic aliases for `aws_` options.

src/metaxy/ext/metadata_stores/_ducklake_support.py
```diff
@@ -1,174 +1,219 @@
...
+
+class S3StorageBackendConfig(BaseModel):
+    """S3 storage backend for DuckLake."""
+
+    type: Literal["s3"]
+    auth: S3AuthConfig | None = None
+    endpoint_url: str | None = None
+    bucket: str | None = None
+    prefix: str | None = None
+    region: str | None = None
```
danielgafni: Needs a Pydantic alias for the `aws_` variant

src/metaxy/ext/metadata_stores/duckdb.py
```diff
@@ -387,10 +342,13 @@ def open(self, mode: AccessMode = "r") -> Iterator[Self]:
         # Call parent context manager to establish connection
         with super().open(mode):
             try:
-                # Configure DuckLake if needed (only on first entry)
-                if self._ducklake_attachment is not None and self._context_depth == 1:
-                    duckdb_conn = self._duckdb_raw_connection()
-                    self._ducklake_attachment.configure(duckdb_conn)
+                if self._context_depth == 1:
+                    # New connection — extensions and DuckLake must be reconfigured
+                    self._extensions_loaded = False
```
danielgafni: I don't see why would we want this. 
Let's just document that extensions are only installed once at the very first `.open()` call. 
Do not reinstall them again.

docs/examples/ducklake.md
```diff
@@ -0,0 +1,67 @@
...
+---
+
+# DuckLake
+
+## Overview
+
+::: metaxy-example source-link
+    example: ducklake
+
+This example demonstrates how to configure [DuckLake](https://ducklake.select/) as a storage layer for the DuckDB metadata store. DuckLake separates metadata from data storage, enabling use of different backends for each.
```
danielgafni: > DuckLake separates metadata from data storage, enabling use of different backends for each.

That's a bit unclear because Metaxy calls "metadata" what DuckLake calls "data", let's change the wording here.

docs/examples/ducklake.md
```diff
@@ -0,0 +1,67 @@
...
+```toml title="metaxy.toml"
+--8<-- "example-ducklake/metaxy.toml"
+```
+<!-- dprint-ignore-end -->
+
+## Walkthrough
+
+The demo script initializes the store from configuration and previews the SQL statements that would be executed:
+
+<!-- dprint-ignore-start -->
```
danielgafni: You can remove all `dprint-ignore-start` comments - I removed the `dprint` Markdown plugin as it was doing more bad than good.

danielgafni (changes requested):
Amazing, I'm almost happy with the PR. 

I guess the last set of comments, mostly for polishing it.

src/metaxy/ext/metadata_stores/_ducklake_support.py
```diff
@@ -1,174 +1,253 @@
...
+        metadata_params = f"METADATA_PATH '', METADATA_PARAMETERS MAP {{'TYPE': 'postgres', 'SECRET': '{secret_name}'}}"
+        return secret_sql, metadata_params
+
+
+class MotherDuckMetadataBackendConfig(BaseModel):
+    """[MotherDuck](https://motherduck.com/)-managed metadata backend for DuckLake."""
+
+    type: Literal["motherduck"]
+    database: str
+    extensions: Sequence[str | ExtensionSpec] = (ExtensionSpec(name="ducklake"),)
```
danielgafni: Does this also need `motherduck` extension?

geoHeil: https://duckdb.org/docs/stable/core_extensions/motherduck yes

danielgafni (changes requested):

danielgafni (approved):
LGTM

PR #1042, opened 2026-03-11
docs: fix ducklake example

geoHeil:


danielgafni:
What are the "historical and deduplicated rows" the example is talking about?

PR #1072, opened 2026-03-18
docs: add example BigQuery E2E

geoHeil:
## Summary

- add docs example for how to use bigquery with alembic
- add workaround example for smooth DDL generation

## Changelog

- add docs example for how to use bigquery with alembic

## Test Plan

- tested with my own BQ account and the included tests


danielgafni (changes requested):
I don't really see general value in this example:

- It's mostly focused on Alembic and not on BigQuery
- The provided Alembic glue code attempting to fix the existing limitations is not a useful documentation artifact, because: 
  1. it's opinionated
  2. it's wrong - for example, it uses `STRUCT` columns with specific field keys which will break the moment these fields are changed in the Metaxy feature 
  3. it's not battle tested

Additionally, the example doesn't have a `.example.yaml` and is not tested in integration tests. 

---
 
I think you should stick to integration tests for BigQuery. There is no need to make them a user-facing example.

Right now it's not obvious to me which aspects of working with BigQuery deserve a dedicated example. What's specific to BigQuery that Metaxy docs don't currently cover?

 The only way to find out is to try it in your real setup, and then it will become more clear over time.

geoHeil:
For the BQ golden tests: it looks like they are fairly slow to pass like 40secs each. should we gate them somehow so they do not always run?

danielgafni (changes requested):
It doesn't seem like you've added `BigQueryMetadataStore` to the shared test suite?

danielgafni (changes requested):

geoHeil:
cleaned up as discussed - secrets are needed now.

Issue #706, opened 2026-01-21
[cli] add `metadata query` command


danielgafni:
It's useful to quickly execute arbitrary SQL against the feature metadata.  This command could enable this. It should run on the result of `read_metadata`. 

For Ibis we should try to do everything in the DB. 

For other backends we can: 

- convert to Polars and run SQL in polars (doesn't have good SQL coverage yet)
- convert to DuckDB (requries an extra dependency but IMO that's fine) 

We can then either print the output or write it to a Parquet file if requires (IMO that's a good extra functionality to have)

PR #1109, opened 2026-03-27
feat(cli): add `metaxy metadata read` command

AnkitSharma-29:
This PR implements the `metadata read` command as requested in #706. It allows users to quickly execute arbitrary SQL queries against feature metadata using DuckDB and Polars. The command supports a simplified interface with `--filter` and `--select` arguments, as well as an advanced `--query` argument. Output can be directed to `stdout` or a file in `markdown`, `json`, `csv`, or `parquet` formats. Comprehensive unit tests have been added in `tests/cli/test_cli_metadata_read.py` to ensure all functionality works as expected.


danielgafni (changes requested):
Thanks! 

Overall looking good, I suggested a few improvements. 

Also, please improve the tests to assert more about the data - ideally the full dataframe for Parquet (json and csv will probably lose proper datatypes)

danielgafni (changes requested):
Left a bunch of comments 

Overall this looks suspiciously AI-ish, please review the output better

danielgafni (changes requested):
1. Minor comments
2. Please consolidate `Literal["csv", "json", "parquet", "markdown"]` into `WriteFormat: TypeAlias = ... ` at the top of this module
3. I think the tests should be refactored: 
    1. pamaterization: instead of defining a single test for every combinations of `format` and `output` parameters, make a single test using [`@pytest.mark.parametrize`](https://docs.pytest.org/en/stable/example/parametrize.html) for both parameters. 
    2. data: I think it should be possible to make a single shared [`@pytest.fixture`](https://docs.pytest.org/en/stable/explanation/fixtures.html) for the datafarme used in these tests. 
4. When `--query` is used with a non-Ibis store, please let's not use Polars at all. Just use `duckdb` instead. We can fail with a clear message if `duckdb` is not available. This is because Polars' support for SQL is way behind DuckDB's. 
5. Since 0.1.7 has been released and Metaxy aims to eventually support `Map` types by default, let's replace all `collect_to_polars` with `_collect_to_arrow`, and use DuckDB instead of Polars for writes. For Markdown we can use the [Markdown extension](https://duckdb.org/community_extensions/extensions/markdown). Avoid using Polars completely, let's do not have the `import polars` at all. This is because Polars doesn't support the `Map` type. 

I am documenting references to all of these docs for educational purposes since previously you said you wanted to learn. I assume good faith here so please take your time to go though them and understand what's expected without AI assistance :) As you can see AI is struggling to write clean (or even correct at all) code without proper human guidance and is useless without it.

src/metaxy/cli/metadata.py
```diff
@@ -731,3 +736,180 @@ def copy(
...
+    elif fmt == "markdown":
+        try:
+            con.install_extension("markdown")
+            con.load_extension("markdown")
+            md = con.query("SELECT * FROM _output").fetchdf().to_markdown(index=False)
+            if md is None:
+                md = ""
+        except Exception:
+            # Fallback: use DuckDB's built-in display
+            md = str(con.query("SELECT * FROM _output"))
```
danielgafni: I think this should just fail with an error. The current fallback doesn't produce Markdown while the user has clearly requested it.

src/metaxy/cli/metadata.py
```diff
@@ -731,3 +736,180 @@ def copy(
...
+        if output:
+            con.execute(f"COPY _output TO '{output}' (FORMAT PARQUET)")
+        else:
+            import pyarrow.parquet as pq
+
+            pq.write_table(table, sys.stdout.buffer)
+    elif fmt == "json":
+        result = con.query("SELECT * FROM _output").fetch_arrow_table()
+        if hasattr(result, "read_all"):
+            result = result.read_all()
```
danielgafni: this check seems a bit weird to me, why do we need to check the attr? `result` has a known type right?

danielgafni (changes requested):
Thanks, two small comments left!

src/metaxy/cli/metadata.py
AnkitSharma-29: You're right, fetch_arrow_table() always returns a pa.Table, so the check was dead code. Removed it here and in the non-Ibis query path where the same pattern appeared.

AnkitSharma-29:
> Thanks, two small comments left!

Thanks! Both addressed in 261bec5.

src/metaxy/cli/metadata.py
AnkitSharma-29: Good catch....removed the fallback entirely so it no longer emits non-markdown output. While fixing it I found the install_extension("markdown") call was the actual cause of the fallback being hit, and it's unnecessary since markdown is rendered via pandas to_markdown(). It now produces real Markdown and fails loudly if rendering can't happen. Added a test covering the failure case.

danielgafni: Please do not use Pandas. Just fix the markdown dudkcb extension installation instead (if that's a thing). Avoid Pandas at all cost.

AnkitSharma-29: Removed pandas from the markdown output it's now generated natively from the DuckDB Arrow result, no to_markdown(). There's no installable DuckDB markdown extension, so the table is built inline. pushed in ddff9f4.

PR #1136, opened 2026-03-31
feat: add `FeatureSpec.unique` field

### LINKED ISSUE #736: [core] Add deduplicate_by field to FeatureSpec
This should enable deduplicating records from datasets managed by Metaxy. 

For example, content-based [semantic hashing](https://github.com/MinishLab/semhash) could be used to identify duplicate samples. Or users could deduplicate by `metaxy_data_version` (given they expect it to be unique, which isn't always true).

geoHeil:
resolves: #736
Related: #1231

## Summary

- add feature-level read-time uniqueness through `FeatureSpec.unique`
- support stable `keep="any"` and deterministic `keep="latest"` through required logical `order_by` columns
- resolve the selected current/lifecycle view before feature uniqueness, with `apply_unique=False` as the explicit raw-view escape hatch
- fail closed when a read requests uniqueness but the feature plan is not registered
- preserve exact projection while loading uniqueness/order columns internally, including Ray datasource parity
- document content-hash value identity separately from durable entity identity

## Semantics

Read order is: selected feature-version scope -> current sample/lifecycle resolution -> `FeatureSpec.unique` -> user filters -> requested projection.

For `keep="latest"`, `order_by` followed by feature ID columns defines the total logical order. NULL sorts below every non-NULL value. Physical system fields break remaining ties. Per-ID current resolution still prioritizes physical recency, so a logical ordinal that must dominate belongs in `id_columns` and then reaches the uniqueness step.

Default reads remove Metaxy-soft-deleted current samples before uniqueness. With `include_soft_deleted=True`, deleted rows participate normally and may win. Domain tombstones remain ordinary ordered rows; downstream filtering belongs on the `FeatureDep` edge.

Feature-history current reads retain one winner per `(id_columns, feature_version)`, preserving the rows needed to reconstruct the current deduplicated view. Raw history for an unregistered feature requires both history flags and `apply_unique=False`.

This provides the deterministic ordered-read and latest-only retention foundation related to #1231. Lifecycle pinning, compaction, conflict validation, snapshot watermarks, changed-key reads, and atomic commits remain separate follow-up work.

## Changelog

- Add `FeatureSpec.unique` with ordered `subset`, `keep`, and required `order_by` for `keep="latest"`.
- Add `MetadataStore.read(..., apply_unique=False)`.
- Serialized specs containing `unique` require a Metaxy version that supports this field.

## Test Plan

- `uv run pytest --tach -q` (1,335 passed, 1 skipped; 1,953 unaffected tests skipped)
- shared uniqueness contract on Delta, DuckDB, DuckLake, PostgreSQL, and ClickHouse (9 passed per backend)
- `uv run pytest -q tests/metadata_stores/core/test_unique.py` (27 passed)
- `uv run pytest -q docs/guide/concepts/definitions/features.md` (12 passed)
- `uv run --all-extras --group dev ty check`
- `uv run ruff check src/metaxy tests`
- `uv run --group docs --extra mcp mkdocs build --clean --strict`

### DIFF
--- a/src/metaxy/metadata_store/base.py
+++ b/src/metaxy/metadata_store/base.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
-from collections.abc import Iterator, Mapping, Sequence
+from collections.abc import Iterable, Iterator, Mapping, Sequence
 from contextlib import AbstractContextManager, contextmanager
 from datetime import datetime, timezone
 from types import TracebackType
@@ -439,7 +439,7 @@ class MetadataStore(ABC):
             *target_filter_list,
         ]
 
-        # Read current metadata with deduplication (with_sample_history=False by default)
+        # Read current metadata with uniqueness resolution (with_sample_history=False by default)
         # Use allow_fallback=False since we only want metadata from THIS store
         # to determine what needs to be updated locally
         try:
@@ -732,6 +732,7 @@ class MetadataStore(ABC):
         with_feature_history: bool = False,
         with_sample_history: bool = False,
         include_soft_deleted: bool = False,
+        apply_unique: bool = True,
         with_store_info: Literal[False] = False,
     ) -> nw.LazyFrame[Any]: ...
 
@@ -747,6 +748,7 @@ class MetadataStore(ABC):
         with_feature_history: bool = False,
         with_sample_history: bool = False,
         include_soft_deleted: bool = False,
+        apply_unique: bool = True,
         with_store_info: Literal[True],
     ) -> tuple[nw.LazyFrame[Any], MetadataStore]: ...
 
@@ -761,6 +763,7 @@ class MetadataStore(ABC):
         with_feature_history: bool = False,
         with_sample_history: bool = False,
         include_soft_deleted: bool = False,
+        apply_unique: bool = True,
         with_store_info: bool = False,
     ) -> nw.LazyFrame[Any] | tuple[nw.LazyFrame[Any], MetadataStore]:
         """
@@ -781,8 +784,12 @@ class MetadataStore(ABC):
             with_feature_history: If True, include rows from all historical feature versions.
                 By default (False), only returns rows with the currently registered feature version.
             with_sample_history: If True, include all historical materializations per sample.
-                By default (False), deduplicates samples within `id_columns` groups ordered by `metaxy_created_at`.
+                By default (False), deduplicates samples within `id_columns` groups ordered by
+                `coalesce(metaxy_deleted_at, metaxy_updated_at)`, keeping the row with the latest timestamp.
             include_soft_deleted: If `True`, include soft-deleted rows in the result. Previous historical materializations of the same feature version will be effectively removed from the output otherwise.
+            apply_unique: If `True`, apply `FeatureSpec.unique` within the selected feature-version
+                scope before sample-history resolution and soft-delete filtering. Set to `False`
+                to disable feature-level uniqueness for this read.
             with_store_info: If `True`, return a tuple of (LazyFrame, MetadataStore) where
                 the MetadataStore is the store that actually contained the feature (which
                 may be a fallback store if allow_fallback=True).
@@ -796,9 +803,10 @@ class MetadataStore(ABC):
             ValueError: If both feature_version and with_feature_history=False are provided
 
         !!! info
-            When this method is called with default arguments, it will return the latest (by `metaxy_created_at`)
-            metadata for the current feature version excluding soft-deleted rows. Therefore, it's perfectly suitable
-            for most use cases.
+            When this method is called with default arguments, it returns the current metadata for the
+            active feature version, excluding soft-deleted rows. The current row for each `id_columns`
+            group is selected by ordering on `coalesce(metaxy_deleted_at, metaxy_updated_at)` and keeping
+            the latest timestamp. Therefore, it's perfectly suitable for most use cases.
 
         !!! warning
             The order of rows is not guaranteed.
@@ -813,6 +821,7 @@ class MetadataStore(ABC):
 
         feature_key = self._resolve_feature_key(feature)
         is_system_table = self._is_system_table(feature_key)
+        feature_plan: FeaturePlan | None = None
 
         # Sync external features if auto-sync is enabled (default)
         # This call is a no-op most of the time and is very lightweight when it's not
@@ -830,9 +839,10 @@ class MetadataStore(ABC):
                 "Use with_feature_history=True with feature_version parameter."
             )
 
-        # Separate system filters (applied before dedup) from user filters (applied after dedup)
+        # Separate system filters (applied before read-mode resolution) from user filters
+        # (applied after read-mode resolution)
         # System filters like feature_version need to be applied early to reduce data volume
-        # User filters should see the deduplicated view of the data
+        # User filters should see the resolved view of the data
         system_filters: list[nw.Expr] = []
         user_filters = list(filters) if filters else []
 
@@ -849,17 +859,28 @@ class MetadataStore(ABC):
         if user_filters:
             read_columns = None
         elif columns and not is_system_table:
-            # Add only system columns that aren't already in the user's columns list
-            columns_set = set(columns)
-            missing_system_cols = [c for c in ALL_SYSTEM_COLUMNS if c not in columns_set]
-            read_columns = [*columns, *missing_system_cols]
+            additional_columns: Iterable[str] | None = None
+            if apply_unique or not with_sample_history:
+                graph = current_graph()
+                definition = graph.feature_definitions_by_key.get(feature_key)
+                if definition is not None:
+                    feature_plan = graph.get_feature_plan(feature_key)
+
+                extra: list[str] = []
+                if not with_sample_history and feature_plan is not None:
+                    extra.extend(feature_plan.feature.id_columns)
+                if apply_unique and feature_plan is not None and feature_plan.feature.unique is not None:
+                    extra.extend(feature_plan.feature.unique.subset)
+                    extra.extend(feature_plan.feature.id_columns)
+                additional_columns = extra or None
+            read_columns = self._internal_read_columns(columns, additional_columns=additional_columns)
         else:
             read_columns = None
 
         lazy_frame = None
         try:
             # Only pass system filters to _read_feature
-            # User filters will be applied after deduplication
+            # User filters will be applied after uniqueness resolution
             lazy_frame = self._read_feature(
                 feature, filters=system_filters if system_filters else None, columns=read_columns
             )
@@ -877,21 +898,27 @@ class MetadataStore(ABC):
             )
 
         if lazy_frame is not None and not is_system_table:
-            # Deduplicate first, then filter soft-deleted rows
-            if not with_sample_history:
-                id_cols = list(self._resolve_feature_plan(feature_key).feature.id_columns)
-                # Treat soft-deletes like hard deletes by ordering on the
-                # most recent lifecycle timestamp.
-                lazy_frame = self.versioning_engine_cls.keep_latest_by_group(
-                    df=lazy_frame,
-                    group_columns=id_cols,
-                    timestamp_columns=[METAXY_DELETED_AT, METAXY_UPDATED_AT],
-                )
+            resolved_feature_plan = feature_plan
+            if resolved_feature_plan is None and (apply_unique or not with_sample_history):
+                graph = current_graph()
+                if feature_key in graph.feature_definitions_by_key:
+                    resolved_feature_plan = graph.get_feature_plan(feature_key)
 
-            if filter_deleted:
+            if apply_unique and resolved_feature_plan is not None:
+                lazy_frame = self._apply_unique(lazy_frame, plan=resolved_feature_plan)
+
+            if not with_sample_history:
+                if resolved_feature_plan is None:
+                    raise RuntimeError(f"Failed to resolve feature plan for {feature_key.to_string()}.")
+                lazy_frame = self._apply_current_feature_view(
+                    lazy_frame,
+                    plan=resolved_feature_plan,
+                    filter_deleted=filter_deleted,
+                )
+            elif filter_deleted:
                 lazy_frame = lazy_frame.filter(nw.col(METAXY_DELETED_AT).is_null())
 
-            # Apply user filters AFTER deduplication so they see the latest version of each row
+            # Apply user filters AFTER read-mode resolution so they see the final row set.
             for user_filter in user_filters:
                 lazy_frame = lazy_frame.filter(user_filter)
 
@@ -926,6 +953,7 @@ class MetadataStore(ABC):
                                 with_feature_history=with_feature_history,
                                 with_sample_history=with_sample_history,
                                 include_soft_deleted=include_soft_deleted,
+                                apply_unique=apply_unique,
                                 with_store_info=True,
                             )
                         return store.read(
@@ -937,6 +965,7 @@ class MetadataStore(ABC):
                             with_feature_history=with_feature_history,
                             with_sample_history=with_sample_history,
                             include_soft_deleted=include_soft_deleted,
+                            apply_unique=apply_unique,
                         )
                 except FeatureNotFoundError:
                     # Try next fallback store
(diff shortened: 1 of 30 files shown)


danielgafni (commented):
I skimmed through the PR and it makes sense! I'd only group `deduplicate_by` and `deduplicate_keep` under a `Deduplication` model, just to keep them more organized. And we could add other kinds of `keep` in other PRs on top of this one. 

One thing I thought of now is perhaps another place to configure deduplication might be `FeatureDep`. I think it should be possible to add this setting to both `FeatureSpec` and `FeatureDep`. Again can be done in another PR. 

I'll let it sit for a while and will think more about this feature. You can refactor the two settings under `Deduplication` Pydantic model in the meanwhile.

src/metaxy/models/feature.py
```diff
@@ -794,6 +835,7 @@ def __new__(
             else:
                 new_cls.__metaxy_project__ = cls._detect_project(new_cls)  # ty: ignore[unresolved-attribute]
 
+            _validate_deduplication_columns(spec, new_cls, active_graph)
```
danielgafni: Looks like this has to be done on `FeatureDefinition` once it's constructed, not in the metaclass.

src/metaxy/models/feature_spec.py
```diff
@@ -202,6 +226,37 @@ def _validate_deps(value: Any) -> list[FeatureDep]:
     return result
 
 
+@public
+class Deduplication(FrozenBaseModel):
+    """Read-time deduplication settings for a feature."""
+
+    by: Annotated[tuple[str, ...], BeforeValidator(_validate_deduplication_by)] = pydantic.Field(
```
danielgafni: note to myself: maybe let's just use `list[str]`, tuples are more annoying to write and don't bring much benefit. I'll think about this again

danielgafni (changes requested):
A bunch of feedback. Didn't look at the tests yet.

src/metaxy/metadata_store/base.py
```diff
@@ -781,8 +781,11 @@ def read(
             with_feature_history: If True, include rows from all historical feature versions.
                 By default (False), only returns rows with the currently registered feature version.
             with_sample_history: If True, include all historical materializations per sample.
-                By default (False), deduplicates samples within `id_columns` groups ordered by `metaxy_created_at`.
+                By default (False), deduplicates samples within `id_columns` groups ordered by
+                `coalesce(metaxy_deleted_at, metaxy_updated_at)`, keeping the row with the latest timestamp.
+                This disables `FeatureSpec.deduplication`, since historical rows are being requested.
```
danielgafni: I am not sure this should be the default behavior.

Deduplication by user columns is an orthogonal option to metadeata history. It should be always enabled even when selecting historical rows.

There should be a separate parameter added to `.read` that could be used to disable deduplication.

src/metaxy/metadata_store/base.py
```diff
@@ -781,8 +781,11 @@ def read(
...
             include_soft_deleted: If `True`, include soft-deleted rows in the result. Previous historical materializations of the same feature version will be effectively removed from the output otherwise.
+                This also disables `FeatureSpec.deduplication`, so deleted rows are not collapsed into active ones.
```
danielgafni: Again I am not sure this should be the default behavior for the same reasons as above.

danielgafni (changes requested):
1. I think the current interaction between custom deduplication and built-in Metaxy's historical overwrites and soft deletes are not ideal. Since deduplication is part of the feature spec, it should always be applied, unless explicitly requires otherwise by the user. It's part of the feature properties. It should be applied before other steps such as historical filtering and soft-deletion filtering.  
2. How about we change the naming to `FeatureSpec.unique` with a `subset` parameter instead of `by`? Sounds more in line with other APIs that are also polars-like. See https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.unique.html#polars.DataFrame.unique

geoHeil:
> 1. I think the current interaction between custom deduplication and built-in Metaxy's historical overwrites and soft deletes are not ideal. Since deduplication is part of the feature spec, it should always be applied, unless explicitly requires otherwise by the user. It's part of the feature properties. It should be applied before other steps such as historical filtering and soft-deletion filtering.
> 
>     2. How about we change the naming to `FeatureSpec.unique` with a `subset` parameter instead of `by`? Sounds more in line with other APIs that are also polars-like. See https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.unique.html#polars.DataFrame.unique

I think this is better so (your approach) - should be addressed now

src/metaxy/metadata_store/base.py
```diff
@@ -1451,6 +1479,61 @@ def _resolve_feature_plan(self, feature: CoercibleToFeatureKey) -> FeaturePlan:
...
+        if unique is None:
+            return lazy_frame
+
+        subset = sorted(unique.subset)
+        if unique.keep == "latest":
+            return self.versioning_engine_cls.keep_latest_by_group(
+                df=lazy_frame,
+                group_columns=subset,
+                timestamp_columns=[METAXY_UPDATED_AT],
+            )
```
danielgafni, line 1565: other variants should handled or an exception should be raised for invalid variants

src/metaxy/metadata_store/base.py
```diff
@@ -1451,6 +1479,61 @@ def _resolve_feature_plan(self, feature: CoercibleToFeatureKey) -> FeaturePlan:
...
+
+        return lazy_frame.unique(subset=subset)
```
danielgafni: we cannot make non-deterministic actions, is this one deterministic according to Narwhals docs? does `.unique` take arguments that can control it?

src/metaxy/models/feature.py
```diff
@@ -116,6 +118,8 @@ def add_feature_definition(
                 f"Each feature key must be unique within a graph."
             )
 
+        _validate_unique_columns(definition.spec, self)
```
danielgafni: this should be done on `FeatureDefinition` itself, not when adding feature definitions to the graph. because it doesn't actually need the graph in any way

src/metaxy/models/feature_spec.py
```diff
@@ -206,11 +207,41 @@ def _validate_deps(value: Any) -> list[FeatureDep]:
     return result
 
 
+@public
+class Unique(FrozenBaseModel):
+    """Read-time uniqueness settings for a feature."""
+
+    subset: frozenset[str] = pydantic.Field(
```
danielgafni: I am pretty sure we want a tuple here since a set would not be ordered?

src/metaxy/metadata_store/base.py
```diff
@@ -1451,6 +1479,61 @@ def _resolve_feature_plan(self, feature: CoercibleToFeatureKey) -> FeaturePlan:
...
+        lazy_frame: nw.LazyFrame[Any],
+        *,
+        plan: FeaturePlan,
+    ) -> nw.LazyFrame[Any]:
+        """Apply read-time uniqueness configured by FeatureSpec.unique."""
+        unique = plan.feature.unique
+        if unique is None:
+            return lazy_frame
+
+        subset = sorted(unique.subset)
```
danielgafni: sorting will reorder the `.subset` which is wrong! it will affect deduplication results. 

We should add a regression test for this bug: the order in which the user has provided `subset` should matter when there are multiple rows to discard for each subset dimension

src/metaxy/models/feature_spec.py
```diff
@@ -206,11 +207,41 @@ def _validate_deps(value: Any) -> list[FeatureDep]:
...
+    )
+
+    if TYPE_CHECKING:
+
+        def __init__(self, *, subset: Sequence[str], keep: UniqueKeep = "any") -> None: ...
+
+    @pydantic.field_serializer("subset")
+    @staticmethod
+    def _serialize_subset(value: frozenset[str]) -> list[str]:
+        return sorted(value)
```
danielgafni: same bug, we should not reorder the user-provided value

src/metaxy/models/feature_spec.py
```diff
@@ -232,6 +263,13 @@ class FeatureSpec(FrozenBaseModel):
         default=None,
         description="Human-readable description of this feature.",
     )
+    unique: Unique | None = pydantic.Field(
+        default=None,
+        description=(
+            "Read-time uniqueness settings. Applied as a feature property before "
+            "sample-history resolution and soft-deletion filtering."
```
danielgafni: can you drop "as a feature property"

danielgafni (changes requested):
ok that's better, but I found a bug that has to be fixed and tested

src/metaxy/models/feature.py
```diff
@@ -763,6 +765,45 @@ class FeatureConfig(TypedDict, total=False):
...
+    """Warn if deduplication.by references missing columns.
+
+    Uses a warning rather than an error because the metadata store allows extra
+    columns beyond model_fields at write time.
+    """
+    deduplication = spec.deduplication
+    if deduplication is None:
+        return
+
+    available = set(feature_cls.model_fields.keys()) | ALL_SYSTEM_COLUMNS
```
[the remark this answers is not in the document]
geoHeil: Resolved. Validation now lives on `FeatureDefinition` and uses `FeatureDefinition.columns`; it makes no assumption that feature classes are Pydantic models.

src/metaxy/models/feature.py
geoHeil: Resolved. The uniqueness-column check is now a `FeatureDefinition` model validator and runs after the definition is constructed.

src/metaxy/models/feature.py
```diff
@@ -763,6 +765,45 @@ class FeatureConfig(TypedDict, total=False):
...
+        # because the resolver implicitly propagates join-required ID columns
+        # and lineage columns even when they are not in select.
+        upstream_def = graph.feature_definitions_by_key.get(dep.feature)
+        if upstream_def is not None:
+            upstream_cols = set(upstream_def.feature_schema.get("properties", {}))
+            if dep.rename:
+                upstream_cols = {dep.rename.get(c, c) for c in upstream_cols}
+            available |= upstream_cols
+        elif dep.select is None:
+            has_unresolved_dep = True
```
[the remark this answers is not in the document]
geoHeil: Resolved. The check now considers only current-definition columns plus Metaxy system columns. Dependency-provided columns remain runtime-compatible but produce a warn-only message because they are absent from the current declared schema.

src/metaxy/models/feature_spec.py
geoHeil: Kept an ordered tuple. `subset` and `order_by` order is preserved in serialization, spec hashing, and backend calls; duplicate column names normalize without sorting, and sets are rejected.

src/metaxy/models/feature_spec.py
```diff
@@ -266,8 +331,42 @@ def __init__(
...
+        if legacy_by is None:
+            if legacy_keep != "any":
+                raise ValueError("deduplicate_keep requires deduplicate_by to be set.")
+            return normalized
+
+        normalized["deduplication"] = {
+            "by": legacy_by,
+            "keep": legacy_keep,
+        }
+        return normalized
```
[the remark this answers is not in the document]
geoHeil: Removed. There is no legacy `deduplicate_*` normalization path; the public configuration surface is only `FeatureSpec.unique`.

src/metaxy/models/feature_spec.py
```diff
@@ -314,6 +413,33 @@ def validate_id_columns(self) -> Self:
...
+            return None
+        return deduplication.by
+
+    @property
+    def deduplicate_keep(self) -> DeduplicationKeep:
+        """Compatibility accessor for legacy deduplicate_keep call sites."""
+        deduplication = self.deduplication
+        if deduplication is None:
+            return "any"
+        return deduplication.keep
```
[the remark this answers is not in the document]
geoHeil: Removed the legacy compatibility accessors.

src/metaxy/models/feature_spec.py
```diff
@@ -202,6 +226,37 @@ def _validate_deps(value: Any) -> list[FeatureDep]:
...
+    def validate_by(self) -> Self:
+        """Validate deduplication columns."""
+        if len(self.by) == 0:
+            raise ValueError("deduplication.by must be non-empty.")
+        for col in self.by:
+            if not col:
+                raise ValueError("deduplication.by must not contain empty strings.")
+        if len(self.by) != len(set(self.by)):
+            raise ValueError("deduplication.by must not contain duplicate column names.")
+        return self
```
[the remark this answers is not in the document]
geoHeil: Used a non-empty ordered tuple rather than `conset`: deterministic order is part of the serialized spec and the backend ordering contract. Bare strings and sets are rejected explicitly.

src/metaxy/models/feature_spec.py
```diff
@@ -176,6 +177,29 @@ def _validate_id_columns(value: Any) -> tuple[str, ...]:
...
+        raise TypeError("deduplicate_by must be a sequence of column names")
+    return tuple(value)
+
+
+def _validate_deduplication_by(value: Any) -> tuple[str, ...]:
+    """Coerce deduplication.by to tuple."""
+    deduplicate_by = _validate_deduplicate_by(value)
+    if deduplicate_by is None:
+        raise TypeError("deduplication.by must be a sequence of column names")
+    return deduplicate_by
```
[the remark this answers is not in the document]
geoHeil: Resolved. `keep` is validated by the Pydantic `Literal`. The remaining sequence validators only reject ambiguous bare strings, mappings, and unordered sets while preserving user order.

src/metaxy/models/feature_spec.py
```diff
@@ -314,6 +413,33 @@ def validate_id_columns(self) -> Self:
...
+    @property
+    def deduplicate_keep(self) -> DeduplicationKeep:
+        """Compatibility accessor for legacy deduplicate_keep call sites."""
+        deduplication = self.deduplication
+        if deduplication is None:
+            return "any"
+        return deduplication.keep
+
+    @pydantic.model_serializer(mode="wrap")
+    def serialize_model(self, handler: pydantic.SerializerFunctionWrapHandler) -> dict[str, Any]:
```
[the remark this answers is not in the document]
geoHeil: Removed the custom model serializer. The field serializers now only preserve ordered tuples as JSON lists.

src/metaxy/metadata_store/base.py
geoHeil: Resolved. `latest` requires an explicit logical `order_by`; feature IDs provide final logical tie-breakers, and scalar system fields only stabilize identical re-appends. Narwhals `unique(..., keep="last", order_by=...)` selects one whole row. Delta and DuckDB/Ibis tests cover reversed input, equal-order ties, history copies, and whole-row pairing.

src/metaxy/metadata_store/base.py
geoHeil: Resolved. Feature uniqueness is default-on for current and historical reads, with the separate `apply_unique=False` escape hatch. Read-mode and soft-delete resolution run first so stale or deleted physical rows cannot erase a valid current duplicate; history uniqueness still applies across the selected visible history.

src/metaxy/models/feature_spec.py
```diff
@@ -206,11 +200,41 @@ def _validate_deps(value: Any) -> list[FeatureDep]:
...
+
 @public
 class FeatureSpec(FrozenBaseModel):
     key: Annotated[FeatureKey, BeforeValidator(FeatureKeyAdapter.validate_python)]
-    id_columns: Annotated[tuple[str, ...], BeforeValidator(_validate_id_columns)] = pydantic.Field(
+    id_columns: tuple[str, ...] = pydantic.Field(
         ...,
+        min_length=1,
         description="Columns that uniquely identify a sample in this feature.",
     )
```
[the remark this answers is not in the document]
geoHeil: Resolved. Bare-string `id_columns` now raises a Pydantic `ValidationError`; ordered sequences are normalized to tuples, preserve order, and remove repeated names. Regression tests cover strings and unordered sets.

src/metaxy/models/feature_spec.py
geoHeil: Resolved. User-provided order is retained without sorting or set conversion, including serialization and feature-spec versioning.

src/metaxy/metadata_store/base.py
geoHeil: Resolved with the same default-on behavior. `MetadataStore.read(..., apply_unique=False)` exposes the unresolved read mode explicitly, and physical maintenance callers use it where required.

src/metaxy/models/feature_spec.py
geoHeil: Agreed. The public fields are ordered tuples, and unordered sets are rejected.

src/metaxy/models/feature.py
geoHeil: Resolved. Validation is now implemented directly on `FeatureDefinition`; graph insertion no longer performs this check.

src/metaxy/models/feature_spec.py
geoHeil: Removed that wording and updated the description to match the implemented read phase ordering.

src/metaxy/metadata_store/base.py
geoHeil: Resolved. The exact `subset` order is preserved in the model, JSON serialization, spec hash, and backend call. Since a permutation defines the same equality groups, winner selection is controlled by `order_by`, not subset position; tests cover both ordered pass-through and deterministic data selection.

geoHeil:
Will pause this here for now as I found a better solution with the expanding relationship for my usecase.

PR #1205, opened 2026-05-04
fix(dagster): table preview for `Map` columns with non-primitive value types

peterroelants:
#### Observations

A Dagster asset materialization in a downstream project produced this error twice (once for the main asset, once for the `__errors` companion) while the materialization itself succeeded:

```
metaxy/ext/dagster/utils.py:578 build_runtime_feature_metadata
  -> metaxy/ext/dagster/table_metadata.py:411 _prepare_dataframe_for_table_record
  -> df.select(exprs)

polars.exceptions.InvalidOperationError: cannot cast List type (inner: 'Int64', to: 'String')

Failing expr: element().struct.field_by_name(value)().strict_cast(String)
Outer expr:   col("encoding_shape").to_physical().list.eval(...)
```

The triggering column was a feature field typed `dict[str, list[int]]`, which metaxy stores as `Map[str, List[Int64]]` (a `polars_map.Map` extension whose physical layout is `List(Struct({key, value}))`).

#### Inferences

- `_map_to_json_expr` was the only path that handled Map columns in the table-preview formatter. It assumed `Map.value` was always castable to `String` (correct for `Map[str, str]`, the only Map columns metaxy emitted before).
- A Map whose value is itself a List/Struct/Array/Map would break the same way; `List[Int64]` was simply the first such type to ship.

### Solution

Encode both the key and the value via Polars' native `json_encode`, which handles every dtype recursively. Polars only exposes `json_encode` on `Struct`, so the value is wrapped in a 1-field struct, encoded, and the wrapper is stripped via regex.

The same trick was already duplicated as a nested `to_json` helper inside `_truncate_list_expr`. Both copies are now replaced by a single module-level `_json_encode_expr`.

Output format:

- `Map[str, str]` → `'{"a": "1", "b": "2"}'` (unchanged — existing snapshot still passes)
- `Map[str, List[Int64]]` → `'{"motion": [12,750,256], "audio": [12,1500]}'`
- Any nested complex value (Struct, Array, Map) is JSON-encoded recursively.

### Changes

- `src/metaxy/ext/dagster/table_metadata.py`
    - New module-level `_json_encode_expr(expr)`: wrap in a 1-field struct, `struct.json_encode()`, strip wrapper via regex.
    - `_map_to_json_expr`: route both key and value through `_json_encode_expr` (replaces the broken `cast(pl.String)` on the value).
    - `_truncate_list_expr`: drop the local `to_json` and call `_json_encode_expr` instead.
- `tests/ext/dagster/test_table_metadata.py`
    - Add `test_build_table_preview_map_with_list_values` — regression test that builds a `Map[str, List[Int64]]` column and asserts the rendered preview. Pre-fix it raised the exact `cannot cast List type` error from the production log.

### Verification

- `ruff check`, `ruff format --check`, `ty check` on touched files: clean.
- `pytest tests/ext/dagster/test_table_metadata.py`: 14 passed, 9 snapshots passed.
- `pytest tests/ext/dagster/`: 215 passed, 72 skipped (clickhouse-marker), 0 failed.
- All three modified helpers are `_`\-prefixed and have no callers outside `table_metadata.py` (verified by grep).

### DIFF
--- a/src/metaxy/ext/dagster/table_metadata.py
+++ b/src/metaxy/ext/dagster/table_metadata.py
@@ -411,6 +411,18 @@ def _prepare_dataframe_for_table_record(df: pl.DataFrame) -> pl.DataFrame:
     return df.select(exprs)
 
 
+def _json_encode_expr(expr: pl.Expr) -> pl.Expr:
+    """Return a Polars Expr that JSON-encodes the value of `expr` to a String.
+
+    Polars exposes `.struct.json_encode()` only on Struct dtype, so to encode
+    a value of arbitrary dtype we wrap it in a 1-field struct, encode that,
+    then strip the wrapper via regex. Output is a JSON literal: strings are
+    quoted, numbers/bools/null are bare, lists/structs/arrays/maps are
+    recursively encoded.
+    """
+    return pl.struct(expr.alias("_")).struct.json_encode().str.extract(r'\{"_":(.*)\}', 1)
+
+
 def _truncate_list_expr(list_expr: pl.Expr, alias: str, max_items: int = 2) -> pl.Expr:
     """Truncate a list expression and convert to string.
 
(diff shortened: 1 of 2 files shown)


peterroelants:
* **#1205** <a href="https://app.graphite.com/github/pr/anam-org/metaxy/1205?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/anam-org/metaxy/1205?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

## How to use the Graphite Merge Queue

Add either label to this PR to merge it via the merge queue:
- <kbd>graphite:merge</kbd> - adds this PR to the back of the merge queue
- <kbd>graphite:hotfix</kbd> - for urgent changes, fast-track this PR to the front of the merge queue

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/anam-org?ref=merge-queue-instructions-in-stack-comment&prId=1205).

<h2></h2>
<sub>

An organization admin has enabled the [Graphite Merge Queue](https://graphite.dev/docs/graphite-merge-queue) in this repository.

</sub>

<sub>

Please do not merge from GitHub as this will restart CI on PRs being processed by the merge queue.

</sub>

<h2></h2>

danielgafni (changes requested):
It looks like the string-casting way of doing this was wrong in the first place. 

I think we should use [`polars.format`](https://docs.pola.rs/api/python/dev/reference/expressions/api/polars.format.html#polars-format) instead. 

I will make a PR to do that. 

---

I checked and it still uses a string cast under the hood 🤦 I'll think about it for a bit... we probably just need proper recursive formatting code

peterroelants:
> It looks like the string-casting way of doing this was wrong in the first place.
> 
> I think we should use [`polars.format`](https://docs.pola.rs/api/python/dev/reference/expressions/api/polars.format.html#polars-format) instead.
> 
> I will make a PR to do that.


Sounds like the right way to go.
Please do close this PR once you have an alternative

